PR #30, opened 2021-04-02
Added code as suggested. Loking forward on how to test

henrj:


appwindow.js
```diff
@@ -55,6 +55,9 @@ var AppWindow = GObject.registerClass(
             this.toggle_action = this.simple_action('toggle', this.toggle.bind(this));
             this.hide_action = this.simple_action('hide', () => this.hide());
 
+            this.toggle_maximize_action = this.simple_action('toggle-maximize', this.toggle_maximize.bind(this));
+            this.unmaximize_action = this.simple_action('unmaximize', () => this.unmaximize());
```
amezin, line 59: Why do you need a dedicated `unmaximize`?

application.js
```diff
@@ -104,6 +104,8 @@ const Application = GObject.registerClass(
 
             this.add_action(this.window.toggle_action);
             this.add_action(this.window.hide_action);
+            this.add_action(this.window.toggle_maximize_action);
+            this.add_action(this.window.toggle_unmaximize_action);
```
amezin, line 108: This copies `win.*` actions to `app.*`. Shouldn't be necessary.

amezin: If I'm not mistaken, this is necessary only for actions that you want to work globally - i. e. when terminal window isn't active. IMO it's not needed for maximize/unmaximize

glade/prefs.ui
```diff
@@ -305,6 +305,12 @@
         <col id="2">0</col>
         <col id="3">0</col>
       </row>
+      <row>
+        <col id="0">ddterm-toggle-maximize-hotkey</col>
+        <col id="1" translatable="yes">Toggle Terminal Window Max Size</col>
```
amezin, line 310: Maybe "Maximize/Unmaximize Terminal Window"?

amezin (changes requested):
I think you missed the step "connect the shortcut to the action".

amezin:
For testing look at https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/docs/INSTALL.md#install-from-git-repository
Just skip the first step (`git clone`)

amezin:
Hm. Actually, the extension tries to unmaximize the window when it's maximized: https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/extension.js#L285 This is because GNOME's window manager tries to auto-maximize the window sometimes (when it covers more than 80% of the screen work area). So you'll have to modify the extension too. I would add `maximize-window` to the settings schema [after `window-height`](https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/schemas/com.github.amezin.ddterm.gschema.xml#L72), change its value from the app, and modify the extension to react to it/take it into account when resizing the window.

henrj:
> For testing look at https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/docs/INSTALL.md#install-from-git-repository
> Just skip the first step (`git clone`)

The extension does not open (Fedora 34)

Tried looking in the log with:
`journalctl /usr/bin/gnome-shell -f -o cat`
Nothing related to ddterm there.

No error logged in looking glass either (`Alt + F2 -> lg -> Enter`)

Thanks

amezin:
1) There is an issue in application code (`journalctl --user -f`):
```
апр 08 13:01:05 trx40 gjs[21139]: JS ERROR: Error: Expected an object of type GAction for argument 'action' but got type undefined
                                     startup@/home/amezin/.local/share/gnome-shell/extensions/ddterm@amezin.github.com/application.js:108:18
                                     @/home/amezin/.local/share/gnome-shell/extensions/ddterm@amezin.github.com/application.js:233:5
```

henrj:
Thanks @amezin, did not think in looking for it there `journalctl --user -f`

I'll look into it tomorrow.

scarlehoff:
I'm very much interested in this feature as well. Let me know if you need more hands for testing (using arch + gnome40)

amezin:
@henrj Do you still plan to work on this?

amezin:
Superseded by https://github.com/amezin/gnome-shell-extension-ddterm/pull/40

Issue #29, opened 2021-04-01
Keyboard shortcut for Maximizing Terminal


henrj:
First of all thanks for this amazing extension, it feels so integrated with the gnome-shell and its a pleasure to use.

The only thing I miss coming from Guake its the feature where a keyboard shortcut toggles the terminal window to fullscreen and back to the original size.

In guake the default its F11.

Was looking at the code to see how can I contribute or at least try to add this feature with no success.

Where should I start looking?

amezin:
1. Add `maximize` `Gio.Action` to `appwindow.js`: https://github.com/amezin/gnome-shell-extension-ddterm/blob/c062f0dc2289c8695c23d822066230d1dfcc2566/appwindow.js#L55-L66
2. Add keyboard shortcut to settings schema: https://github.com/amezin/gnome-shell-extension-ddterm/blob/c062f0dc2289c8695c23d822066230d1dfcc2566/schemas/com.github.amezin.ddterm.gschema.xml#L246-L330
3. Add keyboard shortcut to shortcut table in `prefs.ui`: https://github.com/amezin/gnome-shell-extension-ddterm/blob/c062f0dc2289c8695c23d822066230d1dfcc2566/glade/prefs.ui#L301-L464
4. Connect the shortcut to the action in `application.js`: https://github.com/amezin/gnome-shell-extension-ddterm/blob/c062f0dc2289c8695c23d822066230d1dfcc2566/application.js#L122-L137

henrj:
**Thanks** for the detailed reply... Looking into it

PR #40, opened 2021-05-02
Add a shortcut for toggling the maximed status

scarlehoff:
This is an attempt following the instructions in #29/#30 to add a shortcut that toggles the maximized state

#### Problem
I wanted a method to quickly make the terminal cover the full screen/go back to "half-screen". It should be persistent between F12-presses and between monitors.

#### Attempted solution
When the user presses the shortcut (let's say F11) the screen height is changed to 100% and the original height stored.
If the user presses F11 again, the original height is stored.
On first run if there was an "original height" stored it restores it (and then clean that setting).

One reason for doing it in this way (instead of using for instance `fullscreen`) is resizing after maximizing, which I like doing very much.

#### Testing
I've been using it and seems to work fine. Tested only in Gnome 40 though, but I think I'm not doing anything crazy that should break on any version? Maybe I am... 

Note: first time doing gtk-stuff so there's a bit of cargo cult programming here... let me know if I did something very wrong by gtk standards! In particular, I'm not sure whether by I'm being naughty by doing `set_double("blabla", null)`...

### DIFF
--- a/appwindow.js
+++ b/appwindow.js
@@ -53,26 +53,10 @@ var AppWindow = GObject.registerClass(
             this.toggle_action = this.simple_action('toggle', this.toggle.bind(this));
             this.hide_action = this.simple_action('hide', () => this.hide());
             this.simple_action('toggle-maximize', () => {
-                // Maximize the terminal and store the current height
-                const original_height = this.settings.get_double('original-window-height');
-                let current_height = this.settings.get_double('window-height');
-                let target_value = 1.0;
-
-                if (original_height && current_height === 1.0) {
-                    // if there was an original height _and_ we are full-screen recover it and clean the setting
-                    target_value = original_height;
-                    current_height = null;
-                }
-
-                this.settings.set_double('window-height', target_value);
-                this.settings.set_double('original-window-height', current_height);
+                this.settings.set_boolean('window-maximize', !this.settings.get_boolean('window-maximize'));
             });
-            // On first run, recover the original value if any
-            const original_height = this.settings.get_double('original-window-height');
-            const current_height = this.settings.get_double('window-height');
-            if (original_height && current_height === 1.0)
-                this.settings.set_double('window-height', original_height);
-            this.settings.set_double('original-window-height', null);
+            this.settings.set_boolean('window-maximize', false);
+
 
             this.simple_action('new-tab', this.insert_page.bind(this, -1));
             this.simple_action('new-tab-front', this.insert_page.bind(this, 0));
--- a/appwindow.js
+++ b/appwindow.js
@@ -53,12 +53,27 @@ var AppWindow = GObject.registerClass(
             this.toggle_action = this.simple_action('toggle', this.toggle.bind(this));
             this.hide_action = this.simple_action('hide', () => this.hide());
             this.simple_action('toggle-maximize', () => {
+                // Maximize the terminal and store the current height
+                const original_height = this.settings.get_double("original-window-height");
+                let current_height = this.settings.get_double("window-height");
                 let target_value = 1.0;
-                // If the terminal is already maximized, recover the original value, otherwise maximize it
-                if (this.settings.get_double("window-height") == 1.0) 
-                  target_value = this.settings.get_double("original-window-height");
+
+                if (original_height && current_height == 1.0) {
+                    // if there was an original height _and_ we are full-screen recover it and clean the setting
+                    target_value = original_height;
+                    current_height = null;
+                }
+
                 this.settings.set_double("window-height", target_value);
+                this.settings.set_double("original-window-height", current_height);
             });
+            // On first run, recover the original value if any
+            const original_height = this.settings.get_double("original-window-height");
+            const current_height = this.settings.get_double("window-height");
+            if (original_height && current_height == 1.0) {
+                this.settings.set_double("window-height", original_height);
+            }
+            this.settings.set_double("original-window-height", null);
 
             this.simple_action('new-tab', this.insert_page.bind(this, -1));
             this.simple_action('new-tab-front', this.insert_page.bind(this, 0));
--- a/appwindow.js
+++ b/appwindow.js
@@ -55,7 +55,6 @@ var AppWindow = GObject.registerClass(
             this.simple_action('toggle-maximize', () => {
                 this.settings.set_boolean('window-maximize', !this.settings.get_boolean('window-maximize'));
             });
-            this.settings.set_boolean('window-maximize', false);
 
 
             this.simple_action('new-tab', this.insert_page.bind(this, -1));
--- a/appwindow.js
+++ b/appwindow.js
@@ -52,6 +52,13 @@ var AppWindow = GObject.registerClass(
 
             this.toggle_action = this.simple_action('toggle', this.toggle.bind(this));
             this.hide_action = this.simple_action('hide', () => this.hide());
+            this.simple_action('toggle-maximize', () => {
+                let target_value = 1.0;
+                // If the terminal is already maximized, recover the original value, otherwise maximize it
+                if (this.settings.get_double("window-height") == 1.0) 
+                  target_value = this.settings.get_double("original-window-height");
+                this.settings.set_double("window-height", target_value);
+            });
 
             this.simple_action('new-tab', this.insert_page.bind(this, -1));
             this.simple_action('new-tab-front', this.insert_page.bind(this, 0));
--- a/appwindow.js
+++ b/appwindow.js
@@ -56,7 +56,6 @@ var AppWindow = GObject.registerClass(
                 this.settings.set_boolean('window-maximize', !this.settings.get_boolean('window-maximize'));
             });
 
-
             this.simple_action('new-tab', this.insert_page.bind(this, -1));
             this.simple_action('new-tab-front', this.insert_page.bind(this, 0));
 
--- a/appwindow.js
+++ b/appwindow.js
@@ -54,26 +54,25 @@ var AppWindow = GObject.registerClass(
             this.hide_action = this.simple_action('hide', () => this.hide());
             this.simple_action('toggle-maximize', () => {
                 // Maximize the terminal and store the current height
-                const original_height = this.settings.get_double("original-window-height");
-                let current_height = this.settings.get_double("window-height");
+                const original_height = this.settings.get_double('original-window-height');
+                let current_height = this.settings.get_double('window-height');
                 let target_value = 1.0;
 
-                if (original_height && current_height == 1.0) {
+                if (original_height && current_height === 1.0) {
                     // if there was an original height _and_ we are full-screen recover it and clean the setting
                     target_value = original_height;
                     current_height = null;
                 }
 
-                this.settings.set_double("window-height", target_value);
-                this.settings.set_double("original-window-height", current_height);
+                this.settings.set_double('window-height', target_value);
+                this.settings.set_double('original-window-height', current_height);
             });
             // On first run, recover the original value if any
-            const original_height = this.settings.get_double("original-window-height");
-            const current_height = this.settings.get_double("window-height");
-            if (original_height && current_height == 1.0) {
-                this.settings.set_double("window-height", original_height);
-            }
-            this.settings.set_double("original-window-height", null);
+            const original_height = this.settings.get_double('original-window-height');
+            const current_height = this.settings.get_double('window-height');
+            if (original_height && current_height === 1.0)
+                this.settings.set_double('window-height', original_height);
+            this.settings.set_double('original-window-height', null);
 
             this.simple_action('new-tab', this.insert_page.bind(this, -1));
             this.simple_action('new-tab-front', this.insert_page.bind(this, 0));
--- a/extension.js
+++ b/extension.js
@@ -37,14 +37,10 @@ class ExtensionDBusInterface {
 
         Main.wm.skipNextEffect(current_window.get_compositor_private());
 
-        const was_maximized = settings.get_boolean('window-maximize');
         current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
 
-        if (was_maximized) {
-            // If we started maximized, then cover the whole workarea
-            const workarea = Main.layoutManager.getWorkAreaForMonitor(Main.layoutManager.currentMonitor.index);
-            move_resize_window(current_window, workarea);
-        }
+        const workarea = Main.layoutManager.getWorkAreaForMonitor(Main.layoutManager.currentMonitor.index);
+        move_resize_window(current_window, workarea);
     }
 
     Toggle() {
@@ -116,9 +112,10 @@ function enable() {
 
     settings.connect('changed::window-above', set_window_above);
     settings.connect('changed::window-stick', set_window_stick);
-    settings.connect('changed::window-height', update_window_height);
+    settings.connect('changed::window-height', unset_window_maximize);
+    settings.connect('changed::window-height', update_window_geometry);
     settings.connect('changed::window-skip-taskbar', set_skip_taskbar);
-    settings.connect('changed::window-maximize', set_window_maximize);
+    settings.connect('changed::window-maximize', toggle_window_maximize);
 
     DBUS_INTERFACE.export(Gio.DBus.session, '/org/gnome/Shell/Extensions/ddterm');
 }
@@ -318,8 +315,7 @@ function track_window(win) {
     set_window_above();
     set_window_stick();
     set_skip_taskbar();
-    if (settings.get_boolean('window-maximize') && !win.maximized_vertically)
-        win.maximize(Meta.MaximizeFlags.VERTICAL);
+    toggle_window_maximize();
 }
 
 function workarea_for_window(win) {
@@ -372,28 +368,28 @@ function update_height_setting(win) {
     settings.set_double('window-height', Math.min(1.0, current_height));
 }
 
-function set_window_maximize() {
+function toggle_window_maximize() {
     if (!current_window)
         return;
 
-    const should_maximize = settings.get_boolean('window-maximize');
-    if (current_window.maximized_vertically && !should_maximize)
+    if (current_window.maximized_vertically === settings.get_boolean('window-maximize'))
+        return;
+
+    if (current_window.maximized_vertically)
         current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
-    else if (!current_window.maximized_vertically && should_maximize)
+    else
         current_window.maximize(Meta.MaximizeFlags.VERTICAL);
 }
 
-function update_window_height() {
+function unset_window_maximize() {
     // Wrapper around upadte_window_geometry to ensure that
     // maximize state is always off after a height change
     settings.set_boolean('window-maximize', false);
-    update_window_geometry();
 }
 
 function update_window_geometry() {
     if (!current_window)
         return;
-
 
     const workarea = workarea_for_window(current_window);
     if (!workarea)
@@ -451,8 +447,9 @@ function disconnect_settings() {
     if (settings) {
         GObject.signal_handlers_disconnect_by_func(settings, set_window_above);
         GObject.signal_handlers_disconnect_by_func(settings, set_window_stick);
-        GObject.signal_handlers_disconnect_by_func(settings, update_window_height);
+        GObject.signal_handlers_disconnect_by_func(settings, unset_window_maximize);
+        GObject.signal_handlers_disconnect_by_func(settings, update_window_geometry);
         GObject.signal_handlers_disconnect_by_func(settings, set_skip_taskbar);
-        GObject.signal_handlers_disconnect_by_func(settings, set_window_maximize);
+        GObject.signal_handlers_disconnect_by_func(settings, toggle_window_maximize);
     }
 }
--- a/extension.js
+++ b/extension.js
@@ -225,8 +225,11 @@ function focus_window_changed() {
 
     const win = global.display.focus_window;
     if (win !== null) {
-        if (current_window === win || current_window.is_ancestor_of_transient(win))
+        if (current_window === win || current_window.is_ancestor_of_transient(win)) {
+            if (settings.get_boolean('window-maximize'))
+              window_maximize();
             return;
+        }
     }
 
     if (dbus_action_group)
@@ -366,13 +369,14 @@ function window_maximize() {
         return;
 
     const should_maximize = settings.get_boolean('window-maximize');
-    if (current_window.maximized_vertically && should_maximize)
-        return;
-
-    if (should_maximize)
-        current_window.maximize(Meta.MaximizeFlags.VERTICAL);
-    else
+    if (current_window.maximized_vertically) {
         current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+        settings.set_boolean('window-maximize', false);
+    } else {
+        current_window.maximize(Meta.MaximizeFlags.VERTICAL);
+        settings.set_boolean('window-maximize', true);
+    }
+
 }
 
 function update_window_geometry() {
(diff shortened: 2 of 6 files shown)


amezin:
I still like the approach I described in https://github.com/amezin/gnome-shell-extension-ddterm/pull/30#issuecomment-812814496 more. IMO it would be simpler and more straightforward.

> One reason for doing it in this way (instead of using for instance fullscreen) is resizing after maximizing, which I like doing very much.

Currently, when you set window height to 100%, and then hide & show the window once, on Wayland the window is actually maximized. And you can resize it. So, I think, resizing after maximize should work with the approach I described too

scarlehoff:
I think I didn't understand that approach well (probably from lack knowledge on my part) Do I still have to manually set the height and store the old value to go back with the `maximize-window` approach?

Btw, I've tried using [`maximized()`](https://developer.gnome.org/gtkmm/stable/classGtk_1_1Window.html#a23f2755248ee7aaa39802ccb37bba618) but didn't work.

> On X11 window maximization has been disabled for a long time, you may need to comment out this line to make it work

As far as I've tested this branch seems to work fine both in Gnome Wayland and Gnome Xorg.

amezin:
> I think I didn't understand that approach well (probably from lack knowledge on my part) Do I still have to manually set the height and store the old value to go back with the maximize-window approach?

No, you won't have to, that's the point

> Btw, I've tried using maximized() but didn't work.

You'll need to do that from `extension.js`. I. e. add boolean `window-maximized` to settings, change it from the application. In the extension:
1) maximize/unmaximize `current_window` when the setting changes
2) add `if` that disables existing `.unmaximize()` calls (except `BeginResize()`) when `window-maximized` is set to `true`

This will:
1. Avoid saving/restoring window height
2. Show the correct height in "Preferences" even when the window is maximized
3. Not sure how useful - the window will be maximized from window manager's and toolkit's pov

> As far as I've tested this branch seems to work fine both in Gnome Wayland and Gnome Xorg.

With your current approach it's not necessary. It will be necessary if you try to actually maximize the window, not just resize it to 100%.

amezin:
> On X11 window maximization [has been disabled for a long time](https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/appwindow.js#L132), you may need to comment out [this line](https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/appwindow.js#L39) to make it work

I've removed this code in https://github.com/amezin/gnome-shell-extension-ddterm/commit/1c9af7165a94e880d5196cea6306051321363dea - so maximization on Wayland and X11 should work the same way on `master` branch now

scarlehoff:
Thank you very much for your pointers @amezin, much appreciated the "walk-through"*. Now it works with your approach (much cleaner indeed) and seems to work in all situations I can think of.

It has taken an embarrassingly long time to realized that changes to `extension.js` required to log out/in from Gnome Shell... (is there a way of reloading the shell in wayland?). Incidentally, I think it was also the reason why my previous attempt to using `maximize()` failed... 

*and apologies if I'm taking too much of your time!

appwindow.js
```diff
@@ -52,6 +52,11 @@ var AppWindow = GObject.registerClass(
 
             this.toggle_action = this.simple_action('toggle', this.toggle.bind(this));
             this.hide_action = this.simple_action('hide', () => this.hide());
+            this.simple_action('toggle-maximize', () => {
+                this.settings.set_boolean('window-maximize', !this.settings.get_boolean('window-maximize'));
+            });
+            this.settings.set_boolean('window-maximize', false);
```
amezin, line 58: Why is this (`set_boolean(..., false)`) necessary? What's wrong with keeping maximized state from the previous run? Remember, the application shuts down when you close all tabs. With this code, if you closed all tabs, and then opened the terminal again, it will be unmaximized

extension.js
```diff
@@ -224,8 +225,11 @@ function focus_window_changed() {
 
     const win = global.display.focus_window;
     if (win !== null) {
-        if (current_window === win || current_window.is_ancestor_of_transient(win))
+        if (current_window === win || current_window.is_ancestor_of_transient(win)) {
+            if (settings.get_boolean('window-maximize'))
+                window_maximize();
```
amezin, line 230: It's not the correct place for this. Maybe it accidentally works, but the purpose of this function is hiding the terminal if another application is activated (this behavior can be enabled/disabled in Preferences, off by default).

I think initial `maximize()` should be added at the end of [`track_window()`](https://github.com/amezin/gnome-shell-extension-ddterm/blob/1c9af7165a94e880d5196cea6306051321363dea/extension.js#L311) (where new terminal window is detected and configured). Another possible location is [`update_window_geometry()`](https://github.com/amezin/gnome-shell-extension-ddterm/blob/1c9af7165a94e880d5196cea6306051321363dea/extension.js#L371) (i. e. unmaximize if not `window-maximize`, but also ensure the window is maximized if `window-maximize`)

extension.js
```diff
@@ -356,6 +364,19 @@ function update_height_setting(win) {
     settings.set_double('window-height', Math.min(1.0, current_height));
 }
 
+function window_maximize() {
+    if (!current_window)
+        return;
+
+    if (current_window.maximized_vertically) {
+        current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+        settings.set_boolean('window-maximize', false);
```
amezin, line 373: Why do you need to change `window-maximize` from the extension? IMO the application should control the value, and the extension should only read it (and react to changes)

extension.js
```diff
@@ -333,6 +337,10 @@ function unmaximize_window(win) {
     if (!win.maximized_vertically)
         return;
 
+    const maximized = settings.get_boolean('window-maximize');
+    if (maximized)
```
amezin, line 341: Please simply use `settings.get_boolean('window-maximize')` if its value is used only once (and if the expression isn't too long). When you introduce a variable like this, it makes me expect that it will be used later

extension.js
```diff
@@ -110,6 +110,7 @@ function enable() {
     settings.connect('changed::window-stick', set_window_stick);
     settings.connect('changed::window-height', update_window_geometry);
     settings.connect('changed::window-skip-taskbar', set_skip_taskbar);
+    settings.connect('changed::window-maximize', window_maximize);
```
amezin, line 113: For every `connect()` there should be a matching `disconnect()` (or `GObject.signal_handlers_disconnect_by_func()`, see https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/extension.js#L419)

amezin (changes requested):

amezin:
> is there a way of reloading the shell in wayland?

No, you can do it only in X11 session

extension.js
scarlehoff: Because if you press F11 the value is set to `true`. If you then change the height the value will still be `true` even if it is not maximized.

This ensures the value is synced with the actual state of the window next time F11 is pressed.

The other option would be to set it in update window geometry I guess? (i wasn't sure which made more sense)

appwindow.js
scarlehoff: On a new run my expectation is to return to "vanilla" state. I don't have a very strong preference (maybe it's something for the future, adding an option to control this behaviour)

extension.js
amezin: `update_window_geometry()` is called in response to multiple different events, it won't be correct to set `window-maximize` to `false` there.

You probably should instead set `window-maximize` to `false` in [`unmaximize_window()`](https://github.com/amezin/gnome-shell-extension-ddterm/blob/1c9af7165a94e880d5196cea6306051321363dea/extension.js#L329). It's currently called [in response to changes in window's `maximized-vertically` property](https://github.com/amezin/gnome-shell-extension-ddterm/blob/1c9af7165a94e880d5196cea6306051321363dea/extension.js#L296) only. Setting it to `false` under `if (!win.maximized_vertically)` should be enough to keep in sync I think.

extension.js
```diff
@@ -354,6 +360,17 @@ function update_height_setting(win) {
     const workarea = workarea_for_window(win);
     const current_height = win.get_frame_rect().height / workarea.height;
     settings.set_double('window-height', Math.min(1.0, current_height));
+    settings.set_boolean('window-maximize', false);
```
amezin, line 363: `update_height_setting` is called when resize completes (i. e. when you release the mouse button). I think it's not what you want here. It may be called in response to some other events too, but please don't rely on that.

extension.js
scarlehoff: I've opted for `update_height_settings` in the end with the view that after any event that changes the height of the window the maximize flag should be `false`. I've also tried where you say and I cannot see any difference in behavior. I can leave it there or move it to `unmaximize_window`, as you wish.

extension.js
```diff
@@ -354,6 +360,17 @@ function update_height_setting(win) {
...
+}
+
+function window_maximize() {
+    if (!current_window)
+        return;
+
+    if (current_window.maximized_vertically)
```
amezin, line 370: If `window_maximize` is called in response to `window-maximize` changes, you should read `window-maximize` in it. Otherwise it's not obvious that the value can't become out of sync.

Also, if the function also does `unmaximize`, it shouldn't be called `window_maximize`. Maybe `set_[window_]maximized`, `update_maximized`, `sync_maximized`, etc. (Yes, I know, there are already multiple poorly named functions, let's not add another one)

glade/prefs.ui
```diff
@@ -331,6 +331,12 @@
         <col id="2">0</col>
         <col id="3">0</col>
       </row>
+      <row>
+        <col id="0">shortcut-toggle-maximize</col>
+        <col id="1" translatable="yes">Maximize/unmaximize the Window</col>
```
amezin, line 336: `Maximize/Unmaximize ...` (title case)

amezin (changes requested):
Currently resizing the maximized window doesn't work well on X11. I'm not yet sure why

extension.js
amezin: For example, it's not called (at least, should not be) when you change window height from Preferences dialog

extension.js
scarlehoff: you're right, I'll do "toggle" since it is the name of the option as well.

scarlehoff:
I guess the problem with X is that when you start resizing it goes back to the "original" value. It seems to do that in Wayland as well but only for one frame.
I guess when you start resizing with the mouse the height should move to 100% if you are maximized but I'm don't know where to set that option up.

Also, if in maximized state changing the height from the preference box, the app will not be resized. To me it makes sense (you are in the maximized state regardless of what height you are setting) but from your comments I think you expect the slider should also disable maximized state?

amezin:
> I guess when you start resizing with the mouse the height should move to 100% if you are maximized but I'm don't know where to set that option up.

[`BeginResize`](https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/extension.js#L34) is called when you start dragging the window border. After `unmaximize()` you may either: a) set `window-height` setting to `1.0` and then `update_window_geometry()`, or b) `move_resize_window()` to the entire `workarea` (I haven't tested whether any of these actually solves the problem)

> Also, if in maximized state changing the height from the preference box, the app will not be resized. To me it makes sense (you are in the maximized state regardless of what height you are setting) but from your comments I think you expect the slider should also disable maximized state?

Maybe, I'm not sure. It looks kinda wrong when you drag "Height" slider and nothing changes. Another option is a "Maximized" checkbox, and disabling "Height" slider when maximized

extension.js
amezin: From a `toggle_*()` function I expect that it will invert some boolean (if it was enabled, disable; if it was disabled - enable). Though I'm not an expert in English and may be wrong :) I called similar functions simply `set_*()`: see [`set_window_above`](https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/extension.js#L252), `set_window_stick` - they also set (update) window properties based on boolean settings

scarlehoff:
> BeginResize is called when you start dragging the window border. After unmaximize() you may either: a) set window-height setting to 1.0 a

Yes, this seems to work.

> Maybe, I'm not sure. It looks kinda wrong when you drag "Height" slider and nothing changes. Another option is a "Maximized" checkbox, and disabling "Height" slider when maximized

Do you know whether there's a way to differentiate whether `update_window_geometry` is being called from the slider or not? (so that the `maximize` is ignored when being called from the slider).

extension.js
```diff
@@ -36,6 +36,10 @@ class ExtensionDBusInterface {
             return;
 
         Main.wm.skipNextEffect(current_window.get_compositor_private());
+        if (settings.get_boolean('window-maximize')) {
+            settings.set_boolean('window-maximize', false);
+            settings.set_double('window-height', 1.0);
+        }
```
scarlehoff, line 42: This last commit seems to work well with all "constraints" but I'm wondering whether it would more sense to call here `update_window_height` like:
```suggestion
     update_window_height()
```
so that the same code is not repeated in several places. The problem is I don't know whether there are some conflicts between what `update_window_geometry` and `BeginResize` do.

extension.js
scarlehoff: not my first language either so who am I to say :P I've changed to `set`. It sounds good to me.

extension.js
```diff
@@ -36,6 +36,10 @@ class ExtensionDBusInterface {
             return;
 
         Main.wm.skipNextEffect(current_window.get_compositor_private());
+        if (settings.get_boolean('window-maximize')) {
```
amezin, line 39: IMO this `if` is unnecessary - if the execution reached this point, the window is maximized (look at the `if` at the beginning of the function)

extension.js
```diff
@@ -36,6 +36,10 @@ class ExtensionDBusInterface {
...
+            settings.set_boolean('window-maximize', false);
```
amezin, line 40: Because of `.unmaximize()` call at the end of this function, window's `maximized-vertically` property will be changed to `false`. And that should trigger `unmaximize_window` callback (probably should be renamed to `handle_maximized_changed` or something like that), and it will set `window-maximize` to `false` anyway. So this line should be unnecessary too

extension.js
```diff
@@ -36,6 +36,10 @@ class ExtensionDBusInterface {
...
+            settings.set_double('window-height', 1.0);
```
amezin, line 41: Changing `window-height` will trigger `update_window_height`. Did the other approach (`move_resize_window()` to the entire `workarea`) not work?

And, by the way, resizing should be done after `.unmaximize()` call, I think

extension.js
amezin: See the comments above

amezin (changes requested):

extension.js
scarlehoff: But the fact that the window is maximized vertically is different from whether the user has or not pressed the "maximize window" button, isn't it?

extension.js
scarlehoff: >Changing window-height will trigger update_window_height

You're right! Then it makes more sense to just call `update_window_height()` in this situaton

> Did the other approach (move_resize_window() to the entire workarea) not work?

That's no different from changing the height to 100% right?

> And, by the way, resizing should be done after .unmaximize() call, I think

I think I tried this and the order mattered.

extension.js
```diff
@@ -356,10 +368,31 @@ function update_height_setting(win) {
...
+    if (current_window.maximized_vertically && !should_maximize)
+        current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+    else if (!current_window.maximized_vertically && should_maximize)
+        current_window.maximize(Meta.MaximizeFlags.VERTICAL);
+}
+
+function update_window_height() {
+    if (settings.get_boolean('window-maximize')) {
+        settings.set_boolean('window-maximize', false);
+        settings.set_double('window-height', 1.0);
```
amezin, line 385: Is setting `window-height` really necessary here?

BTW, it will trigger another `update_window_height()` if `window-height` isn't `1.0` already (not sure if later from event loop, or recursively).

extension.js
```diff
@@ -370,7 +403,7 @@ function update_window_geometry() {
 
     if (current_window.maximized_vertically && target_rect.height < workarea.height) {
         Main.wm.skipNextEffect(current_window.get_compositor_private());
-        current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+        unmaximize_window(current_window);
```
amezin, line 406: `unmaximize_window()` expects to be called as `notify::maximized-vertically` handler - that's why it accepts the window as an argument and validates it. Please just add `window-maximized` check into the surrounding `if` and call `.unmaximize` directly.

Also, this code currently will call `Main.wm.skipNextEffect()` but won't `.unmaximize()` if `window-maximized` is `true`. So it will incorrectly disable an unrelated effect.

amezin (changes requested):

scarlehoff:
Setting the size to 1.0 is necessary because otherwise when you start resizing it (with the mouse) it goes back to whatever the original size was. I don't think it can be avoided at all with "this approach" (without even uglier hacks).
Note that the callback from setting the height to 1.0 is called only once.

Basically what I am trying to accomplish (with the latest commits) is the following: if the terminal is maximized and you try to change the size -> unmaximize the terminal so the size changes happen in realtime -> move the size to 100% because that's effectively where the terminal is if you start resizing from a maximized state.

In X11 not doing it like this creates the glitch that we noticed earlier. In Wayland instead even with no hack the window stays where the mouse is.

If I have time later today I'll rewind back a bit and try with `move_resize_window` to see whether the hacks can be avoided and it works across X11 and Wayland without glitches.

I understand now why almost no app allows you to resize when maximized :laughing:  although it is a feature I would like to have.

amezin:
> Setting the size to 1.0 is necessary because otherwise when you start resizing it (with the mouse) it goes back to whatever the original size was. I don't think it can be avoided at all with "this approach" (without even uglier hacks).

But `update_window_height` should not be called when resizing with mouse (only after resizing ends - i. e. when you released the mouse button)

scarlehoff:
The last commit uses `move_resize_window`. Indeed that works. I think I've removed all the hacky stuff that went around the "height to 1.0" change. 

Apologies, I should've tested this one rather than assuming it would be equivalent to setting the height to 1.

extension.js
```diff
@@ -36,7 +36,15 @@ class ExtensionDBusInterface {
             return;
 
         Main.wm.skipNextEffect(current_window.get_compositor_private());
+
+        const was_maximized = settings.get_boolean('window-maximize');
         current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+
+        if (was_maximized) {
```
amezin, line 43: I still think this `if` is not necessary. Window was maximized, mouse resizing starts - it means that the window should occupy the entire workarea, regardless of `window-maximize`

extension.js
```diff
@@ -356,10 +372,29 @@ function update_height_setting(win) {
     settings.set_double('window-height', Math.min(1.0, current_height));
 }
 
+function set_window_maximize() {
+    if (!current_window)
+        return;
+
+    const should_maximize = settings.get_boolean('window-maximize');
+    if (current_window.maximized_vertically && !should_maximize)
```
amezin, line 380: You can simplify things by `if (current_window.maximized_vertically === should_maximize) return;`. And then just `if (should_maximize) maximize(); else unmaximize();`

extension.js
```diff
@@ -309,6 +318,8 @@ function track_window(win) {
     set_window_above();
     set_window_stick();
     set_skip_taskbar();
+    if (settings.get_boolean('window-maximize') && !win.maximized_vertically)
+        win.maximize(Meta.MaximizeFlags.VERTICAL);
```
amezin, line 322: I'm not 100% sure, but maybe you can simply call `set_window_maximize()` here now?

extension.js
```diff
@@ -356,10 +372,29 @@ function update_height_setting(win) {
...
+        return;
+
+    const should_maximize = settings.get_boolean('window-maximize');
+    if (current_window.maximized_vertically && !should_maximize)
+        current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+    else if (!current_window.maximized_vertically && should_maximize)
+        current_window.maximize(Meta.MaximizeFlags.VERTICAL);
+}
+
+function update_window_height() {
```
amezin, line 386: I don't like the ambiguous function name here - it's not clear what's the difference with `update_window_geometry`. Maybe `unmaximize_and_update_window_geometry()` or something like that.

Or you could create a function `unset_window_maximize()`/`disable_window_maximize()` and `.connect()` it to `changed::window-height` before `update_window_geometry()` (don't forget `disconnect()` too).

Actually I would prefer the 2nd option

extension.js
```diff
@@ -356,10 +372,29 @@ function update_height_setting(win) {
...
+    // maximize state is always off after a height change
+    settings.set_boolean('window-maximize', false);
+    update_window_geometry();
+}
+
 function update_window_geometry() {
     if (!current_window)
         return;
 
+
```
amezin, line 397: Please remove this empty line

amezin (changes requested):

amezin:
Overall it works (I'm using this branch now), just some style changes/clarity improvements are needed IMO.

Also, would be nice if you squash all the commits.

Then I'll merge it

appwindow.js
```diff
@@ -52,6 +52,10 @@ var AppWindow = GObject.registerClass(
 
             this.toggle_action = this.simple_action('toggle', this.toggle.bind(this));
             this.hide_action = this.simple_action('hide', () => this.hide());
+            this.simple_action('toggle-maximize', () => {
+                this.settings.set_boolean('window-maximize', !this.settings.get_boolean('window-maximize'));
+            });
+
```
amezin, line 58: Another unnecessary empty line

extension.js
amezin: Yep, `set_window_maximize()` seems to work fine here

scarlehoff:
One last thing, I've changed `set_window_maximize` to `toggle_window_maximize` since otherwise there would be `set` and `unset` `_window_maximize` doing wildly different things.

If you are happy with the style/terminology I'll do the squash.

amezin (changes requested):

extension.js
```diff
@@ -356,6 +368,25 @@ function update_height_setting(win) {
     settings.set_double('window-height', Math.min(1.0, current_height));
 }
 
+function toggle_window_maximize() {
```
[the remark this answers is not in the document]
scarlehoff: Maybe yes. 

Tbh, I wanted to change `unset_window_maximize` but couldn't come up with anything. Maybe `unset_window_maximize` / `set_window_maximized` is a good compromise? Otherwise don't know what to use.

extension.js
```diff
@@ -356,6 +367,26 @@ function update_height_setting(win) {
...
+        return;
+
+    if (should_maximize)
+        current_window.maximize(Meta.MaximizeFlags.VERTICAL);
+    else
+        current_window.unmaximize(Meta.MaximizeFlags.VERTICAL);
+}
+
+function unset_window_maximize() {
+    // Wrapper around upadte_window_geometry to ensure that
```
scarlehoff, line 385:
```suggestion
```

extension.js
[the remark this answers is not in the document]
amezin: `disable_window_maximize_setting` ?

amezin (approved):

amezin:
Merged, thank you!

There was another small issue on older GNOME versions (<40): https://github.com/amezin/gnome-shell-extension-ddterm/issues/41

I would appreciate if you test the current "master" branch for a few days on the GNOME version you use and report if there are any issues.

scarlehoff:
Sure, I will.

amezin:
@scarlehoff I've added copyright+license notices as recommended in https://www.gnu.org/licenses/gpl-howto.html

You could add your "copyright (c)" line in [extension.js](https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/extension.js#L2) too, if you want to, via pull request. Or simply give me a "copyright (c)" line that I can copy and paste

scarlehoff:
I'm happy to "forfeit"/pass copyright to you (nice of you to think of me but I'm just happy I contributed, I feel you deserve all rights on this program!)

If you'd feel bad without adding me you can put me as "juacrumar" (I guess: Copyright © 2021 Juacrumar)

Thanks.

PR #53, opened 2021-06-20
Support using colors from gnome terminal 

ccat3z:
Since a lot of Gnome Shell users use Gnome Terminal, I think it is useful to sync with the color profile of Gnome Terminal.

https://user-images.githubusercontent.com/7223991/122678837-fec15d80-d21a-11eb-9549-5d9f8ac6eb15.mp4

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -57,6 +57,37 @@ function format_scale_value_percent(scale, value) {
     return PERCENT_FORMAT.format(value);
 }
 
+function show_dialog(parent_window, message, message_type = Gtk.MessageType.ERROR) {
+    const dialog = new Gtk.MessageDialog({
+        transient_for: parent_window,
+        modal: true,
+        buttons: Gtk.ButtonsType.CLOSE,
+        message_type,
+        text: message,
+    });
+    dialog.connect('response', () => dialog.destroy());
+    dialog.show();
+}
+
+// Looks up a `Gio.SettingsSchema` with the identifier `schema_id`.
+// Throw error if the schema does not exist.
+function get_settings_schema(schema_id) {
+    const schema_source = Gio.SettingsSchemaSource.get_default();
+    const schema = schema_source.lookup(schema_id, true);
+    if (schema === null)
+        throw new Error(`Settings schema '${schema_id}' not found.`);
+    return schema;
+}
+
+// Get key named `name` from `schema`.
+// Throw error if the key does not exist.
+function get_settings_schema_key(schema, name) {
+    if (!schema.has_key(name))
+        throw new Error(`Key '${name}' does not exist in settings schema '${schema.get_id()}'.`);
+
+    return schema.get_key(name);
+}
+
 function createPrefsWidgetClass(resource_path, util) {
     const cls = GObject.registerClass(
         {
@@ -91,6 +122,7 @@ function createPrefsWidgetClass(resource_path, util) {
                 'bold_color_check',
                 'color_scheme_editor',
                 'color_scheme_combo',
+                'copy_gnome_terminal_profile_button',
                 'palette_combo',
                 'theme_variant_combo',
                 'tab_policy_combo',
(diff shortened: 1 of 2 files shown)


amezin:
Can this be just a button in preferences dialog (i. e. "Copy settings from GNOME Terminal")?

Unchecking the checkbox won't revert settings to the old values, as far as I can tell

ccat3z:
You're right, I rarely change colors profile. I added the "Copy profile from GNOME Terminal" button at the bottom of the colors panel.

prefs.js
```diff
@@ -455,6 +458,45 @@ function createPrefsWidgetClass(resource_path, util) {
...
                 );
             }
+
+            copy_gnome_terminal_profile() {
+                // Find default gnome terminal profile
+                let profiles_list;
+                try {
+                    profiles_list = Gio.Settings.new('org.gnome.Terminal.ProfilesList');
+                } catch (e) {
+                    logError(`Cannot find profiles of gnome terminal: ${e}`);
```
amezin, line 468: This should probably be a message dialog (and the other `logError` too)

amezin (changes requested):
Everything except `logError()`s looks good.

Would be nice if it was possible to select the profile to import (button with a popup menu?) - but it is not a requirement.

prefs.js
```diff
@@ -455,6 +458,50 @@ function createPrefsWidgetClass(resource_path, util) {
...
+                    'bold-color-same-as-fg', 'bold-color',
+                    'cursor-colors-set', 'cursor-foreground-color', 'cursor-background-color',
+                    'highlight-colors-set', 'highlight-foreground-color', 'highlight-background-color',
+                    'palette',
+                    'bold-is-bright',
+                ].forEach(key => {
+                    this.settings.set_value(
+                        key,
+                        gnome_terminal_profile.get_value(key)
+                    );
```
amezin, line 502: `.get_value()`: [It is a programmer error to give a key that isn't contained in the
schema](https://gjs-docs.gnome.org/gio20~2.66p/gio.settings#method-get_value)

`.set_value()`: [It is a programmer error to give a key that isn't contained in the
schema for this or for value to have the incorrect type](https://gjs-docs.gnome.org/gio20~2.66p/gio.settings#method-set_value)

"It is a programmer error" means the application may crash (like `Gio.Settings.new` with nonexistent schema), even if it currently doesn't. And GNOME Terminal's schemas may change.

So I think it's necessary to check that a) all keys exist in the profile schema b) values have expected (by ddterm settings schema) type.

Yep, I've missed multiple issues in the initial review

amezin (changes requested):

util.js
```diff
@@ -41,6 +41,37 @@ function enum_from_settings(nick, enum_class) {
...
+}
+
+// Get key named `name` from `schema`.
+// Throw error if the key does not exist.
+function get_settings_schema_key(schema, name) {
+    if (!schema.has_key(name))
+        throw new Error(`Key '${name}' does not exist in settings schema '${schema.get_id()}'.`);
+
+    return schema.get_key(name);
+}
```
amezin, line 73: Please keep all these functions in `prefs.js` - currently they aren't used by anything outside `prefs.js` (and probably won't ever be).

prefs.js
```diff
@@ -545,6 +548,59 @@ function createPrefsWidgetClass(resource_path, util) {
...
+                        'bold-is-bright',
+                    ];
+
+                    // Check if key is valid
+                    for (const key of profile_keys) {
+                        const type_gnome_terimnal = util.get_settings_schema_key(profile_schema, key).get_value_type();
+                        const type_ddterm = util.get_settings_schema_key(this.settings.settings_schema, key).get_value_type();
+
+                        if (!type_gnome_terimnal.equal(type_ddterm))
+                            throw new Error(`The type of key '${key}' in GNOME Terminal is not as expected.`);
```
amezin, line 591: Would be nice if the message contained the actual types - see [GLib.VariantType.dup_string()](https://gjs-docs.gnome.org/glib20~2.66.1/glib.varianttype#method-dup_string). This would make debugging easier if something ever goes wrong.

amezin (changes requested):
Looks good, only 2 cosmetic changes are necessary, and then I'll merge it. Also, please squash commits into one (I can do it during merge, but your signature will probably be lost this way). Thank you!

amezin (approved):

PR #72, opened 2021-08-08
Add shortcuts for height and opacity

### LINKED ISSUE #70: Shortcut for increasing/decreasing terminal height
This was a nice feature on Guake, with the default binds as `Ctrl + Up` and `Ctrl + Down`.

jacksongoode:
Default keybindings for increasing/decreasing size are take from Guake, opacity bindings are unset.

This would close #70.

### DIFF
--- a/appwindow.js
+++ b/appwindow.js
@@ -107,6 +107,26 @@ var AppWindow = GObject.registerClass(
             this.method_handler(this.settings, 'changed::window-position', this.update_resize_boxes);
             this.update_resize_boxes();
 
+            const HEIGHT_MOD = 0.05;
+            this.simple_action('window-size-dec', () => {
+                if (this.settings.get_boolean('window-maximize'))
+                    this.settings.set_double('window-size', 1.0 - HEIGHT_MOD);
+                else
+                    this.adjust_double_setting('window-size', -HEIGHT_MOD);
+            });
+            this.simple_action('window-size-inc', () => {
+                if (!this.settings.get_boolean('window-maximize'))
+                    this.adjust_double_setting('window-size', HEIGHT_MOD);
+            });
+
+            const OPACITY_MOD = 0.05;
+            this.simple_action('background-opacity-dec', () => {
+                this.adjust_double_setting('background-opacity', -OPACITY_MOD);
+            });
+            this.simple_action('background-opacity-inc', () => {
+                this.adjust_double_setting('background-opacity', OPACITY_MOD);
+            });
+
             this.tab_select_action = new Gio.PropertyAction({
                 name: 'switch-to-tab',
                 object: this.notebook,
@@ -178,6 +198,12 @@ var AppWindow = GObject.registerClass(
             this.signal_connect(action, 'activate', func);
             this.add_action(action);
             return action;
+        }
+
+        adjust_double_setting(name, difference, min = 0.0, max = 1.0) {
+            const current = this.settings.get_double(name);
+            const new_setting = current + difference;
+            this.settings.set_double(name, Math.min(Math.max(new_setting, min), max));
         }
 
         update_tab_bar_visibility() {
--- a/glade/prefs.ui
+++ b/glade/prefs.ui
@@ -498,6 +498,30 @@ along with ddterm GNOME Shell extension.  If not, see <http://www.gnu.org/licens
         <col id="3">0</col>
       </row>
       <row>
+        <col id="0">shortcut-window-size-inc</col>
+        <col id="1" translatable="yes">Increase Window Height</col>
+        <col id="2">0</col>
+        <col id="3">0</col>
+      </row>
+      <row>
+        <col id="0">shortcut-window-size-dec</col>
+        <col id="1" translatable="yes">Decrease Window Height</col>
+        <col id="2">0</col>
+        <col id="3">0</col>
+      </row>
+      <row>
+        <col id="0">shortcut-background-opacity-inc</col>
+        <col id="1" translatable="yes">Increase Background Opacity</col>
+        <col id="2">0</col>
+        <col id="3">0</col>
+      </row>
+      <row>
+        <col id="0">shortcut-background-opacity-dec</col>
+        <col id="1" translatable="yes">Decrease Background Opacity</col>
+        <col id="2">0</col>
+        <col id="3">0</col>
+      </row>
+      <row>
         <col id="0">shortcut-toggle-transparent-background</col>
         <col id="1" translatable="yes">Enable/Disable Background Transparency</col>
         <col id="2">0</col>
--- a/application.js
+++ b/application.js
@@ -143,6 +143,10 @@ const Application = GObject.registerClass(
             this.update_theme();
 
             this.setup_shortcut('shortcut-window-hide', 'win.hide');
+            this.setup_shortcut('shortcut-window-size-inc', 'win.window-size-inc');
+            this.setup_shortcut('shortcut-window-size-dec', 'win.window-size-dec');
+            this.setup_shortcut('shortcut-background-opacity-inc', 'win.background-opacity-inc');
+            this.setup_shortcut('shortcut-background-opacity-dec', 'win.background-opacity-dec');
             this.setup_shortcut('shortcut-toggle-maximize', 'app.window-maximize');
             this.setup_shortcut('shortcut-toggle-transparent-background', 'app.transparent-background');
             this.setup_shortcut('shortcut-terminal-copy', 'terminal.copy');
--- a/schemas/com.github.amezin.ddterm.gschema.xml
+++ b/schemas/com.github.amezin.ddterm.gschema.xml
@@ -376,6 +376,18 @@
     <key name="shortcut-window-hide" type="as">
       <default><![CDATA[[]]]></default>
     </key>
+    <key name="shortcut-window-size-inc" type="as">
+      <default><![CDATA[['<Ctrl>Down']]]></default>
+    </key>
+    <key name="shortcut-window-size-dec" type="as">
+      <default><![CDATA[['<Ctrl>Up']]]></default>
+    </key>
+    <key name="shortcut-background-opacity-inc" type="as">
+      <default><![CDATA[[]]]></default>
+    </key>
+    <key name="shortcut-background-opacity-dec" type="as">
+      <default><![CDATA[[]]]></default>
+    </key>
     <key name="shortcut-toggle-maximize" type="as">
       <default><![CDATA[['F11']]]></default>
     </key>


appwindow.js
```diff
@@ -107,6 +107,40 @@ var AppWindow = GObject.registerClass(
             this.method_handler(this.settings, 'changed::window-position', this.update_resize_boxes);
             this.update_resize_boxes();
 
+            this.simple_action('window-add-height', () => {
+                const current = this.settings.get_double('window-size');
+
+                if (current > 0.95)
+                    this.settings.set_double('window-size', 1.0);
+                else
+                    this.settings.set_double('window-size', current + 0.05);
```
amezin, line 116: I think `0.05` should be a named constant

glade/prefs.ui
```diff
@@ -497,6 +497,30 @@ along with ddterm GNOME Shell extension.  If not, see <http://www.gnu.org/licens
         <col id="2">0</col>
         <col id="3">0</col>
       </row>
+      <row>
+        <col id="0">shortcut-window-add-height</col>
+        <col id="1" translatable="yes">Increase Height of Window</col>
```
amezin, line 502: I think "Increase Window Height" would be better (because it's shorter)

appwindow.js
```diff
@@ -107,6 +107,40 @@ var AppWindow = GObject.registerClass(
...
+                else
+                    this.settings.set_double('background-opacity', current + 0.05);
+            });
+            this.simple_action('window-sub-opacity', () => {
+                const current = this.settings.get_double('background-opacity');
+
+                if (current < 0.05)
+                    this.settings.set_double('background-opacity', 0.0);
+                else
+                    this.settings.set_double('background-opacity', current - 0.05);
```
amezin, line 141: Could you extract this into a function? `adjust_double_setting(name, difference)`. And then use it to implement all 4 actions

amezin (changes requested):
In general, the implementation is correct. Please fix setting and action names, and try to get rid of code duplication.

After that:
1) Please don't change package-lock.json
2) Squash all commits into one

appwindow.js
jacksongoode: Yes, I think it also might make sense to add optional `max` and `min` parameters as well to this.

However, I am running into an issue with this:
`this.simple_action('window-inc-opacity', this.adjust_double_setting('background-opacity', opacity_mod));`

As the `signal_connect` seems to want a function - I can just wrap it like:
`this.simple_action('window-inc-opacity', () => {this.adjust_double_setting('background-opacity', opacity_mod)});`

But there is likely a better solution...

jacksongoode:
Almost done, but it seems there is still a lingering error that's appearing now:

`Can't update stage views actor MetaSurfaceActorX11 is on because it needs an allocation.`

This appears when using the hotkey to increase/decrease the size (silent on transparency changes).

appwindow.js
amezin: > As the `signal_connect` seems to want a function - I can just wrap it like:
> `this.simple_action('window-inc-opacity', () => {this.adjust_double_setting('background-opacity', opacity_mod)});`

Totally acceptable solution (`opacity_mod` is a constant, I guess? If so, please use `UPPER_CASE`).

> Yes, I think it also might make sense to add optional `max` and `min` parameters as well to this.

If I'm not mistaken, min=0 and max=1 for both settings, so these parameters aren't currently necessary. If/when they become necessary, they could be extracted from the settings schema.

amezin:
> Almost done, but it seems there are errors that's appearing now:
> 
> ```
> Can't update stage views actor MetaSurfaceActorX11 is on because it needs an allocation.
> Can't update stage views actor MetaWindowActorX11 is on because it needs an allocation.
> Can't update stage views actor MetaWindowGroup is on because it needs an allocation.
> ```
> 
> This appears everytime the hotkey is used to increase/decrease the size (silent on transparency changes).

You may ignore these warnings for now. Most likely your code isn't responsible for them.

Do they appear if you open preferences and drag the "Window size" slider?

appwindow.js
jacksongoode: Ok great, and the min and max are necessary for the assignment (don't want to increase/substract past range).

How would we extract these props from the schema?

jacksongoode:
> > Almost done, but it seems there are errors that's appearing now:
> > ```
> > Can't update stage views actor MetaSurfaceActorX11 is on because it needs an allocation.
> > Can't update stage views actor MetaWindowActorX11 is on because it needs an allocation.
> > Can't update stage views actor MetaWindowGroup is on because it needs an allocation.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This appears everytime the hotkey is used to increase/decrease the size (silent on transparency changes).
> 
> You may ignore these warnings for now. Most likely your code isn't responsible for them.
> 
> Do they appear if you open preferences and drag the "Window size" slider?

Yes, I will ignore them.

One final issue I've noticed is that changes to `background-opacity` if it is initially set to 1.0 will not appear until the window is moved or selected (focused). Like:

1. Set transparency to 1.0
2. Hide/close window
3. Open window
4. Change transparency in settings 
5. Window does not change until focused by selection or move

I'll file this as a separate bug.

jacksongoode (commented):
Looks good to me so far.

appwindow.js
amezin: https://gjs-docs.gnome.org/gio20~2.66p/gio.settings#property-settings_schema
https://gjs-docs.gnome.org/gio20~2.66p/gio.settingsschema#method-get_key
https://gjs-docs.gnome.org/gio20~2.66p/gio.settingsschemakey#method-get_range

However, the docs say that:
> This information should not be used by normal programs. It is
> considered to be a hint for introspection purposes. Normal programs
> should already know what is permitted by their own schema. The
> format may change in any way in the future -- but particularly, new
> forms may be added to the possibilities described above.

So maybe it's not a good idea

appwindow.js
```diff
@@ -257,8 +284,7 @@ var AppWindow = GObject.registerClass(
         }
 
         update_app_paintable() {
-            this.app_paintable = this.settings.get_boolean('transparent-background') &&
-                                 this.settings.get_double('background-opacity') < 1.0;
```
amezin, line 261: Don't remove this line, it's correct. There is no point in doing custom painting if `background-opacity == 1.0`. If the window doesn't immediately become transparent after changing from `1.0`, try calling [`this.queue_draw()`](https://gjs-docs.gnome.org/gtk20~2.24.33/gtk.widget#method-queue_draw)

appwindow.js
```diff
@@ -107,6 +107,26 @@ var AppWindow = GObject.registerClass(
...
             this.update_resize_boxes();
 
+            const HEIGHT_MOD = 0.05;
+            this.simple_action('window-inc-height', () => {
+                this.adjust_double_setting('window-size', HEIGHT_MOD);
+            });
+            this.simple_action('window-dec-height', () => {
+                const current = this.settings.get_double('window-size');
+                if (current < HEIGHT_MOD)
+                    this.hide();
```
amezin, line 117: Don't do this. The hotkey is for decreasing window size, not hiding the window. `window-size = 0` is fine; you can set it to `0` in the preferences dialog. The window will just have the smallest size possible.

amezin (changes requested):

appwindow.js
jacksongoode: Got it!

appwindow.js
jacksongoode: Ok, then I will file the bug I mentioned above in a separate issue.

jacksongoode:
Not sure about the test for gnome-wayland-nested-dual-monitor - is there something I can do @amezin?

amezin (changes requested):
If `window-maximize` is `true`, resizing should start from `1.0`.

1) Maximize the window (F11 by default)
2) Try to decrease its height.

It will resize to the original (unmaximized) size minus `HEIGHT_MOD`.

IMO it should resize to `1.0 - HEIGHT_MOD` instead.

amezin:
> Not sure about the test for gnome-wayland-nested-dual-monitor - is there something I can do @amezin?

No, the test failure is unrelated to your changes. Tests rely on various timeouts too much, and some configurations thus fail from time to time.

glade/prefs.ui
```diff
@@ -497,6 +497,30 @@ along with ddterm GNOME Shell extension.  If not, see <http://www.gnu.org/licens
...
+      </row>
+      <row>
+        <col id="0">shortcut-window-dec-height</col>
+        <col id="1" translatable="yes">Decrease Window Height</col>
+        <col id="2">0</col>
+        <col id="3">0</col>
+      </row>
+      <row>
+        <col id="0">shortcut-window-inc-opacity</col>
+        <col id="1" translatable="yes">Increase Window Opacity</col>
```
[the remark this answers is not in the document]
amezin: Another nitpick at action/settings names: `inc`/`dec` should probably be a prefix/suffix (i. e. `shortcut-inc-window-size`, `shortcut-dec-background-opacity`; or `shortcut-window-size-inc`, `shortcut-background-opacity-dec`), instead of being inserted in the middle of the affected setting name.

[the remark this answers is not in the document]
jacksongoode: Good points, just saw the issue with the maximized state. Added it in latest commit.

appwindow.js
```diff
@@ -107,6 +107,24 @@ var AppWindow = GObject.registerClass(
             this.method_handler(this.settings, 'changed::window-position', this.update_resize_boxes);
             this.update_resize_boxes();
 
+            const HEIGHT_MOD = 0.05;
+            this.simple_action('window-size-dec', () => {
+                if (this.settings.get_boolean('window-maximize'))
+                    this.settings.set_double('window-size', 1.0);
```
amezin, line 113: You're calling `settings.set_double` twice: here, then in `adjust_double_setting` again. It triggers all callbacks subscribed to `changed::window-size`. Currently, they do nothing after the first `set_double` call, but it might change. Please change the code to call `set_double` only once (i. e. `adjust_double_setting` should not be called if `window-maximize` is `true`)

appwindow.js
```diff
@@ -107,6 +107,24 @@ var AppWindow = GObject.registerClass(
...
+                this.adjust_double_setting('window-size', -HEIGHT_MOD);
+            });
+            this.simple_action('window-size-inc', () => {
+                this.adjust_double_setting('window-size', HEIGHT_MOD);
```
amezin, line 117: This too doesn't work correctly if `window-maximize` is `true` and `window-size` < `0.95`. Simply doing nothing if `window-maximize` is `true` should be fine

amezin (changes requested):

appwindow.js
jacksongoode: Actually there is an outstanding issue with this. I tried to record this but actually the recording shows the proper function!?

Can you duplicate this:
1. Set transparency to 1.0
2. Hide/close window
3. Open window
4. Change transparency in settings
5. Type - you should see a weird ghosting effect on the letters, or some fully white blocks
6. Window does not refreshed until size is changed

amezin: Please report it as a separate issue.

Have you tried adding `this.queue_draw()` to `update_app_paintable()`?

jacksongoode: I have tried queue_draw() to no success, we can continue the conversations here #73.

PR #88, opened 2021-10-05
Add keyboard shortcuts for reordering current tab

mjnaderi:
Thanks for this great extension.

This PR adds two keyboard shortcuts for reordering current tab. One for moving the current tab backward, and one for moving it forward. (Like tilda's "Move Tab Right" and "Move Tab Left" shortcuts).


amezin:
1. Don't worry about failed tests - they are not related to your code (I really need to make them less flaky...)
2. Overall the change looks good
3. I'm not sure about the wording. "Backward" feels wrong here (although I may be wrong)
4. Could you also add corresponding items to the tab's context menu (`menus.ui`, under `<menu id="tab-popup">`)?

mjnaderi:
Hi @amezin. Thanks for your response.

I'm not sure about wording too. I just didn't want to use "Left" and "Right" like tilda, because when tab bar is on right or left edge, left and right don't make sense. I appreciate suggestions.

I added reorder items to context menu, and fixed a bug about updating shortcut labels.

amezin:
@mjnaderi after discussing it with people who know English better than me, I think you should change "backward[s]" to "back" here. I. e. "move the tab back" or "move one tab back"/"move a tab back" (and for "forward" you should add "the" or "one tab"/"a tab" too)

PR #91, opened 2021-10-11
Add `truncate-chars` handlebars helper

mjnaderi:
Using this helper, user can set a maximum length on tab titles (like tilda `Title Max Length` option in tilda).

Without this option, tabs can get too large:

image

Using `{{truncate-chars window-title 20}}`:

image


mjnaderi:
@amezin I updated the example and squashed the commits.

PR #108, opened 2021-12-24
Load custom user style

mjnaderi:
Hi.

I needed to customize styles, so I added the ability to load custom user styles (from `~/.ddterm/style.css`).


amezin:
1) Don't worry about failing tests; they're unrelated to your pull request
2) I like the idea but don't like the hardcoded filename. Possible options:
  a) Store file name in settings (and add file selector to preferences dialog)
  b) Store the stylesheet itself in settings (and add a text area + maybe load/save buttons to preferences dialog - probably worth a dedicated page)

amezin:
Actually, I'd prefer option (b), to keep application settings self-contained.

mjnaderi:
Thanks for your feedback @amezin 
Option (b) was my first idea. I was thinking of adding an "Advanced Options" tab (for custom styles and any future advanced option).
I'm not expert in GTK and gnome extension development and it's a bit hard for me to implement.

amezin:
I doubt you care after 5 years, but maybe someone else will be searching for an answer here.

ddterm's main window now (since v63.2.0) has `#ddterm-window` css id, and custom styles can be applied to it and to its child widgets directly in `~/.config/gtk-3.0/gtk.css`.

PR #114, opened 2022-01-18
Update default bindings to be consistent with other gnome projects

blennster:
Other gnome applications such as gnome-terminal use `Ctrl+Shift+t` to open a new tab and `Ctrl+Shift+w` to close tabs. Nautilus uses `Ctrl+t` and `Ctrl+w` and so do most browsers. I think it would be nice to align with these commands since most people will have learned this and it would be nice if this carried over to this terminal also.

### DIFF
--- a/schemas/com.github.amezin.ddterm.gschema.xml
+++ b/schemas/com.github.amezin.ddterm.gschema.xml
@@ -419,7 +419,7 @@
       <default><![CDATA[[]]]></default>
     </key>
     <key name="shortcut-win-new-tab" type="as">
-      <default><![CDATA[['<Ctrl><Shift>n']]]></default>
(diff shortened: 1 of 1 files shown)


amezin:
These defaults have been there for more than a year. Not sure if it's a good idea to change them now. The people who use them likely got used to it.

Initially, I've used "new window"/"close window" shortcuts because there is only one window. And, if I'm not mistaken, GNOME Terminal's "new tab"/"close tab" shortcuts were conflicting with something.

blennster:
Yeah that's valid. I think it would be nice to get some input from more people but I'm not sure how to go about doing that.

vorot93:
I support changing the default to match the rest of GNOME.

PR #163, opened 2022-08-15
FR PO: non-breaking spaces before colons

arnaud-feldmann:
As asked on https://github.com/ddterm/gnome-shell-extension-ddterm/issues/159#issuecomment-1212372261


amezin:
Thank you!

Issue #79, opened 2021-09-01
No such interface “org.freedesktop.DBus.Properties”


bialad:
Thank you for a great extension!

I'm trying to get the number of opened tabs from the d-bus interface, and looked into the GetAll property, but I get this error. Is it not implemented?

`gdbus call --session --dest org.gnome.Shell --object-path /org/gnome/Shell/Extensions/ddterm --method org.freedesktop.DBus.Properties.GetAll`

**Error: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.DBus.Properties” on object at path /org/gnome/Shell/Extensions/ddterm**

amezin:
The d-bus interface has almost nothing in it: https://github.com/amezin/gnome-shell-extension-ddterm/blob/master/com.github.amezin.ddterm.Extension.xml
No properties at all.
Also, it's the interface exported by the extension, not the application...

In short, yep, it's not implemented

bialad:
Ah, I see. Is it possible to get the opened tabs through bash? I was hoping a call to GetAll properties would return a list of tabs. :)

amezin:
No, it's not possible. If it were, it would be implemented as a D-Bus call.

Why do you need to count open tabs?

bialad:
My idea was to attach a remote tmux session to each tab, so that each tab is the same server every time. Just a convenient feature to have.

If I were to implement this, I could use the ExtensionDBusInterface class, and get the tabs from current_window?

amezin:
You could use ExtensionDBusInterface as an example only. You'll need to design and implement a new interface. The extension doesn't have access to open tabs (or any other window internals), it can only move/resize the window. Look at appwindow.js

PR #297, opened 2023-01-08
extension: add a desktop entry

### LINKED ISSUE #123: Add an icon to ddterm
When ddterm is configured to show up in the Overview and Panel, it shows up with the default cog app icon:
| Overview | Panel |
| ----------- | ----------- |
| image | image |
| Icon in the Overview | Symbolic icon in the panel |

I took a look at how you setup the nice window title ("Drop Down Terminal") and found this snippet [here](https://github.com/amezin/gnome-shell-extension-ddterm/blob/07b383a64ac63ba53b17331f39e207e517b82e31/application.js#L780):

```javascript
GLib.set_prgname('com.github.amezin.ddterm');
GLib.set_application_name('Drop Down Terminal');
Gdk.set_allowed_backends('x11');
```

But it doesn't look there's anything similar in `GLib` to setup app icons. In any case, I think it would be really nice to have a terminal icon for ddterm (perhaps there's a way to have a full blown .desktop file for it, like GSConnect has?).

PS: thanks a lot for coming back to ddterm, it is my favourite GNOME Extension :rocket::heart:

pesader:
Add a desktop file so ddterm is displayed more nicely (with an icon and a proper name) in the activities view and application menu. Install the file to the user's local applications directory when the extension is enabled and remove it when the extension is disabled.

Here's a demo of it working:

https://user-images.githubusercontent.com/65264536/211216070-84468242-4283-4f68-b942-ca8189652f50.mp4

Closes #123.


pesader:
I ran the linter locally and it seems that the only files that are throwing errors are `rxjs.js` and `handlebars.js`, which I did not edit.

amezin:
> I ran the linter locally and it seems that the only files that are throwing errors are `rxjs.js` and `handlebars.js`, which I did not edit.

It shouldn't lint rxjs and handlebars at all - they should be ignored. Try running `make lint`.

Here eslint fails on `extension.js`: https://github.com/ddterm/gnome-shell-extension-ddterm/actions/runs/3868698513/jobs/6594312408#step:7:12

amezin:
I wonder why the error doesn't generate an annotation on the pull request though...

ddterm/shell/extension.js
```diff
@@ -140,6 +140,28 @@ class ExtensionDBusInterface {
     }
 }
 
+class DesktopEntry {
+    static file = Gio.File.new_for_path(
+        GLib.build_filenamev([GLib.get_user_data_dir(), 'applications', `${APP_ID}.desktop`])
+    )
```
amezin, line 146: Could you make everything non-static, create the instance in `enable()`, and set the reference to `null` in `disable()`? Or convert `file()` to a function/property?

https://gjs.guide/extensions/review-guidelines/review-guidelines.html#destroy-all-objects

amezin (changes requested):

ddterm/com.github.amezin.ddterm.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=Drop Down Terminal
+NoDisplay=true
+Comment=Another drop down terminal extension for GNOME Shell. With tabs. Works on Wayland natively.
+Icon=terminal
+Exec=gdbus call --session --dest org.gnome.Shell --object-path /org/gnome/Shell/Extensions/ddterm --method com.github.amezin.ddterm.Extension.Toggle
```
amezin, line 7: `Activate` method might be more appropriate here

ddterm/shell/extension.js
pesader: Of course! Thanks for link to the docs :)

ddterm/com.github.amezin.ddterm.desktop
pesader: From what I read in the [Freedesktop Spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html), the line `DBusActivatable=true` is needed to signal that ddterm can be activated via dbus. Besides the `Activate` method, they suggest implementing an `Open` method for opening files with the application and an `ActivateAction` for desktop actions. Since the last two are not relevant for ddterm, I think we're good without them.

The `Exec` line must be kept anyway, to keep compatibility with implementations that don't handle the `DBusActivatable` line properly.

amezin: No, I mean `Activate` method from ddterm extension's D-Bus interface. `Toggle` hides ddterm if it's visible. `Activate` always shows/brings the window to the front.

No need for D-Bus activation (it can't be correctly implemented here anyway).

pesader:
Thank you for taking the time to review this PR and (especially) for pointing me to good references and learning resources. I've made the changes you requested, but it is now failing a ton of tests on every OS.

I'll try to figure out what's going on.

ddterm/com.github.amezin.ddterm.desktop
pesader: Oh, of course 🤦🏻‍♂️
Sorry, I misunderstood you.

amezin:
There is a lot of noise, but I believe the root cause is that the extension fails to load:

```
[  531.628757] gnome-shell[79]: JS ERROR: Extension ddterm@amezin.github.com: Gio.IOErrorEnum: Error opening file “/home/gnomeshell/.local/share/applications/com.github.amezin.ddterm.desktop”: No such file or directory
install@/home/gnomeshell/.local/share/gnome-shell/extensions/ddterm@amezin.github.com/ddterm/shell/extension.js:166:16
enable@/home/gnomeshell/.local/share/gnome-shell/extensions/ddterm@amezin.github.com/ddterm/shell/extension.js:256:19
enable@/home/gnomeshell/.local/share/gnome-shell/extensions/ddterm@amezin.github.com/extension.js:32:10
_callExtensionEnable@resource:///org/gnome/shell/ui/extensionSystem.js:182:32
_onEnabledExtensionsChanged/<@resource:///org/gnome/shell/ui/extensionSystem.js:527:35
_onEnabledExtensionsChanged@resource:///org/gnome/shell/ui/extensionSystem.js:527:14
createCheckedMethod/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:533:46
enableExtension@resource:///org/gnome/shell/ui/extensionSystem.js:209:29
EnableExtension@resource:///org/gnome/shell/ui/shellDBus.js:447:38
_handleMethodCall@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:310:38
_wrapJSObject/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:387:34
```

https://github.com/ddterm/gnome-shell-extension-ddterm/actions/runs/3905791923/jobs/6673238419#step:13:11571

Maybe `~/.local/share/applications` doesn't exist?

Every concurrent test session spawns a new container, so it shouldn't be a race between file creation/deletion/existence check.

ddterm/shell/extension.js
```diff
@@ -140,6 +142,39 @@ class ExtensionDBusInterface {
...
+                GLib.get_user_data_dir(),
+                'applications',
+                `${APP_ID}.desktop`,
+            ]
+        ));
+    }
+
+    install() {
+        try {
+            Gio.File.new_for_path(GLib.build_filenamev(
```
amezin, line 158: Nitpick: maybe you could create this Gio.File in constructor too? Less indentation -> easier to read

ddterm/com.github.amezin.ddterm.desktop
```diff
@@ -0,0 +1,11 @@
...
+Actions=
+Categories=Utility;
+DBusActivatable=true
```
amezin, line 10: If I'm not mistaken, to support D-Bus activation the application must own a well-known name and it must have a service description file: https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-starting-services

And `org.freedesktop.Application` isn't implemented by the extension either https://specifications.freedesktop.org/desktop-entry-spec/1.1/ar01s07.html

ddterm inherits `org.freedesktop.Application` implementation from GLib/GApplication... But ddterm app shouldn't be started directly.

Although, maybe it's possible to also install a D-Bus service description file, in the same way as the .desktop file (does D-Bus have per-user service directories?). And write the service description in a way that launches ddterm app correctly (i. e. through the extension).

But I'd recommend to experiment with D-Bus activation in a separate pull request anyway.

ddterm/shell/extension.js
```diff
@@ -140,6 +142,39 @@ class ExtensionDBusInterface {
...
+                [
+                    GLib.get_user_data_dir(),
```
amezin, line 160: Oh, there's another issue... The extension can also be installed system-wide. In this case, it won't be stored in user's directory. Please find a way to get the actual source directory. For example, look at how D-Bus interface XML is loaded.

ddterm/com.github.amezin.ddterm.desktop
pesader: Yep, that's definitely a story for another day 😄
But I like the idea, can investigate further later!

ddterm/shell/extension.js
pesader: Done :)

pesader:
It now ensures the user applications directory exists before it installs the desktop file there. I'll now address the case that the extension is installed system-wide.

ddterm/shell/extension.js
pesader: Yep! In addition to that, the right place to put the desktop file also won't be in the user's `$HOME` but on `/usr/share/applications`. Fortunately, there's `GLib.get_system_data_dirs` for that.

Will work on this next!

ddterm/shell/extension.js
```diff
@@ -140,6 +142,50 @@ class ExtensionDBusInterface {
...
+                'ddterm',
+                `${APP_ID}.desktop`,
+            ]
+        ));
+    }
+
+    _ensure_install_directory_exists() {
+        var install_directory = this.target_file.get_parent();
+        if (!GLib.file_test(install_directory.get_path(), GLib.FileTest.IS_DIR))
+            install_directory.make_directory_with_parents(null);
```
amezin, line 169: Could you maybe use [`GLib.mkdir_with_parents`](https://gjs-docs.gnome.org/glib20/glib.mkdir_with_parents) that doesn't throw? And thus avoid checking if the directory already exists

ddterm/com.github.amezin.ddterm.desktop
amezin: If D-Bus activation isn't supported, I think `DBusActivatable=true` should be removed

ddterm/com.github.amezin.ddterm.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Type=Application
+Name=Drop Down Terminal
+NoDisplay=true
+Comment=Another drop down terminal extension for GNOME Shell. With tabs. Works on Wayland natively.
+Icon=terminal
+Exec=gdbus call --session --dest org.gnome.Shell --object-path /org/gnome/Shell/Extensions/ddterm --method com.github.amezin.ddterm.Extension.Activate
+Actions=
```
amezin, line 8: Is there any reason to include empty `Actions=`? I. e. is the default non-empty somehow?

pesader:
All tests passed 🙌🏻 

> Maybe ~/.local/share/applications doesn't exist?

Your intuition was correct!

ddterm/shell/extension.js
```diff
@@ -140,6 +142,50 @@ class ExtensionDBusInterface {
...
+    }
+
+    _is_installed() {
+        return GLib.file_test(this.target_file.get_path(), GLib.FileTest.EXISTS);
```
amezin, line 173: Since `this.target_file` is a `Gio.File`, wouldn't it be easier to call [`.exists()`](https://gjs-docs.gnome.org/gio20~2.0/gio.file#method-query_exists)?

ddterm/shell/extension.js
```diff
@@ -140,6 +142,50 @@ class ExtensionDBusInterface {
...
+    }
+
+    install() {
+        if (!this._is_installed()) {
```
amezin, line 177: Quote from the documentation for [`.exists()`](https://gjs-docs.gnome.org/gio20~2.0/gio.file#method-query_exists):

> Note that in many cases it is [racy to first check for file existence](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use)
> and then execute something based on the outcome of that, because the
> file might have been created or removed in between the operations. The
> general approach to handling that is to not check, but just do the
> operation and handle the errors as they come.

Maybe it would be easier to handle the error from `.copy()` (checking that the error is `EXISTS`, of course)? Or.. does it actually throw an error when the target file already exists?

Another concern: if a previous version didn't clean up properly, an outdated file will be kept...

ddterm/shell/extension.js
```diff
@@ -140,6 +142,50 @@ class ExtensionDBusInterface {
...
+            this._ensure_install_directory_exists();
+            this.source_file.copy(this.target_file, Gio.FileCopyFlags.NONE, null, null);
+        }
+    }
+
+    uninstall() {
+        if (this._is_installed())
+            this.target_file.delete(null);
```
amezin, line 185: Again, maybe it's both easier and more reliable to just handle the error (IMO if the file wasn't in place, it's an error too - I'd just catch and log all errors here, without re-throwing)

amezin (changes requested):
Sorry that I didn't respond with all questions at once - I didn't notice multiple things at the first sight

ddterm/com.github.amezin.ddterm.desktop
pesader: Opsie, I thought I had deleted the commit that made that change. Fixed now.

ddterm/com.github.amezin.ddterm.desktop
pesader: I generated this file with a GUI tool, called [MenuLibre](https://flathub.org/apps/details/org.bluesabre.MenuLibre). It probably leaves it there. I thought it had some importance but now that I tested it without `Actions=` I see it doesn't. Line removed!

ddterm/shell/extension.js
pesader: I'll handle the error as you suggested, but instead of doing nothing if the file already exists I can overwrite the file with `GLib.File.replace()`. Would that solve the issue?

amezin: https://gjs-docs.gnome.org/gio20~2.0/gio.file#method-copy

> If the flag #G_FILE_COPY_OVERWRITE is specified an already
existing destination file is overwritten.

Won't `.copy()` with `Gio.FileCopyFlags.OVERWRITE` solve all issues with just one function call, without needing any checks?

ddterm/shell/extension.js
amezin: You shouldn't need `GLib.get_system_data_dirs` or `GLib.get_user_data_dir`. Again, look at how D-Bus interface XML is loaded, in the same `extension.js`.

pesader: Thanks, I'll use the same method to get the `source_file` that is the extension's directory like so:

```js       Me.dir.get_child('ddterm').get_child('com.github.amezin.ddterm.Extension.xml');
```

However, the `target_file` also changes location when the extension is installed system-wide. It shouldn't be installed in the user's local data directory, but rather the system data directory. For that, I think I'll need `GLib.get_system_data_dirs`.

amezin: `Me.dir` is the extension's source directory. If the extension is installed system-wide, it's `/usr/share/gnome-shell/extensions/ddterm@...` (or whatever gnome-shell's prefix is), when installed in the "usual" way, in user's home, it is `~/.local/share/gnome-shell/extensions/ddterm@...`. There's no need for any additional logic.

ddterm/shell/extension.js
pesader: It works every time using this flag. Thanks for pointing me to that!
Implemented on [5744f79](https://github.com/ddterm/gnome-shell-extension-ddterm/pull/297/commits/5744f79d92c0ebe210b0365b715bc11da11ce39f).

ddterm/shell/extension.js
pesader: I'm using the `OVERWRITE` flag on install and catching errors on uninstall, as suggested, so the _in_installed() function is not needed anymore.

ddterm/shell/extension.js
pesader: That's much better indeed. Fixed in [5744f79](https://github.com/ddterm/gnome-shell-extension-ddterm/pull/297/commits/5744f79d92c0ebe210b0365b715bc11da11ce39f).

pesader:
There seems to be a permission error now, which I don't get.

When using `make_directory_with_parents()` instead of `mkdir_with_parents()` the file permission was 644 (everyone allowed to read, user allowed to read an write), but there's a Permission Error when a test attempts to open the desktop file in the CI now.

ddterm/shell/extension.js
pesader: Got it now! Thanks for taking the time to explain.

ddterm/shell/extension.js
```diff
@@ -140,6 +142,33 @@ class ExtensionDBusInterface {
...
+        GLib.mkdir_with_parents(this.target_file.get_parent().get_path(), 0o700);
+        this.source_file.copy(this.target_file, Gio.FileCopyFlags.OVERWRITE, null, null);
+    }
+
+    uninstall() {
+        try {
+            this.target_file.delete(null);
+        } catch (e) {
+            if (!e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND))
+                throw e;
```
amezin, line 167: Throwing here is a bad idea, because it will interrupt `disable()`.

Could you just log the error here, and not re-throw it?

And, again, if the file is not found here, it means something unexpected happened - I think it should be an error too (i. e. should be logged as an error).

amezin (changes requested):

ddterm/shell/extension.js
pesader: I was going to work on this next, had to take a break to do something else 😅
 
Thanks for reminding me though. Next time, if you prefer, I can request a review when I think it is worth another shot, so you don't lose too much time reviewing my WIPs.

amezin:
Thanks for your patience and fixing all issues

pesader:
Thanks for _your_ patience teaching me how to fix them! Cheers

amezin:
If you wish to work on D-Bus activation - you would really help a lot if you implemented it. Multiple features have it as a prerequisite.

pesader:
Sure would like to :)
Is it issue #79 or should I open another issue to discuss that?

amezin:
@pesader no, that issue, even though related, isn't the correct one.

I've opened a dedicated issue for D-Bus activation: https://github.com/ddterm/gnome-shell-extension-ddterm/issues/304

PR #317, opened 2023-01-30
panelicon: use symbolic icon

pesader:
Append `-symbolic` to the panel icon name, to make sure the symbolic variant is used. Although, this seems not to be needed for the default Adwaita icon theme, it is so for third-party icon themes.

Here's a comparison, using Papirus icon pack:
| Before | After |
|--------|-------|
| Screenshot from 2023-01-29 20-45-15 | Screenshot from 2023-01-29 20-46-45| 

The `utilities-terminal-symbolic` icon is also available on the Adwaita icon theme, so this change does not break the default experience as you can see below.

image

### DIFF
--- a/ddterm/shell/panelicon.js
+++ b/ddterm/shell/panelicon.js
@@ -55,7 +55,7 @@ const PanelIconBase = GObject.registerClass(
             this.connections = new ConnectionSet();
 
             this.add_actor(new St.Icon({
-                icon_name: 'utilities-terminal-symbolic',
+                icon_name: 'terminal-symbolic',
                 style_class: 'system-status-icon',
             }));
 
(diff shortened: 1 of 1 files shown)


pesader:
First Arch failed, now Ubuntu.

This seems like it will create more problems than it solves, so I'm closing it.

amezin:
These tests sometimes randomly fail. If it's one or two tests, usually it's not a real failure. At least, not worse than it was before in `master` branch.

However, I'm not sure if simply replacing the icon name is correct. Maybe the same result can be achieved by adjusting `style_class`?

pesader:
> Or you may make the icon name configurable. "Panel icon" settings page is almost empty anyway.

I like this idea! I'll open a new issue to track that.

Issue #304, opened 2023-01-14
D-Bus Activation


amezin:
Even when ddterm extension is active and installed, ddterm app might not be running (the app shuts down after the last tab is closed). But most D-Bus calls should end up in the application process. `com.github.amezin.ddterm` bus name is owned by the application process too.

D-Bus has an "activation" feaure - D-Bus daemon can automatically launch processes when a call is made to a specific name. 

However, ddterm app should be started only from the extension, especially on Wayland - otherwise some features won't work. [`com.github.amezin.ddterm`](https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/com.github.amezin.ddterm) script shouldn't be called directly. Instead, a D-Bus call to the extension should be performed - similar to the [`.desktop` file](https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/ddterm/com.github.amezin.ddterm.desktop#L7)

The extension should install a D-Bus service description file, similar to the `.desktop` file.

This will allow adding D-Bus features/interfaces to the application, without the need to "proxy" them through the extension, and the users won't have to worry whether the app is actually running. [GJS application packaging spec](https://gjs.guide/guides/gtk/application-packaging.html#index) also suggests applications should be D-Bus activatable.

amezin:
https://dbus.freedesktop.org/doc/dbus-daemon.1.html

If the docs are correct, `$XDG_RUNTIME_DIR/dbus-1/services` is among the directories where D-Bus daemon looks for service files. `$XDG_RUNTIME_DIR` on my machine resolves to `/run/user/1000`. I think it is the most appropriate place for ddterm's service file, since it's generated at runtime, and doesn't need to be stored persistently.

Hm, maybe there's a similar subdir in `$XDG_RUNTIME_DIR` for desktop files...

pesader:
Hi again! I could contribute making the `DesktopFile` class more generic (`InstallableResource` perhaps) so we can use the same code for installing the dbus service file.

We would only need to initialize each `InstallableResource` object with different values for `source_file` and `target_file`.

What do you think?

pesader:
Now on the content of the service file itself, I found an example by GSConnect:

```
[D-BUS Service]
Name=org.gnome.Shell.Extensions.GSConnect
Exec=@PACKAGE_DATADIR@/service/daemon.js
```

[...]

PR #326, opened 2023-02-01
dbus activation: add dbus service file

pesader:
First, refactor the `DesktopEntry` class into a more generic `InstallableResource` class, that doesn't have pre-defined values for `source_file` and `target_file`, and can be used for both the desktop entry file and the dbus service file.

Then, create an instance of this new class for the dbus file and install it to a subdirectory of the user's runtime directory $XDG_RUNTIME_DIR, where the dbus daemon looks for these files.

If merged, this would solve issue #304.

### DIFF
--- a/ddterm/org.gnome.Shell.Extensions.ddterm.service
+++ b/ddterm/org.gnome.Shell.Extensions.ddterm.service
@@ -1,0 +1,3 @@
+[D-BUS Service]
+Name=org.gnome.Shell.Extensions.ddterm
+Exec=gdbus call --session --dest org.gnome.Shell --object-path /org/gnome/Shell/Extensions/ddterm --method com.github.amezin.ddterm.Extension.Activate
--- a/ddterm/shell/extension.js
+++ b/ddterm/shell/extension.js
@@ -262,6 +262,19 @@ function enable() {
             ]))
     );
     dbus_service.install();
+
+    Gio.DBus.session.call(
+        'org.freedesktop.DBus',
+        'org/freedesktop/DBus',
+        'org.freedesktop.DBus',
+        'ReloadConfig',
+        null,
+        null,
+        Gio.DBusCallFlags.NONE,
+        -1,
+        null,
+        null,
+    )
 }
 
 function disable() {
--- a/ddterm/shell/extension.js
+++ b/ddterm/shell/extension.js
@@ -44,6 +44,7 @@ let window_connections = null;
 let dbus_interface = null;
 
 let desktop_entry = null;
+let dbus_service = null;
 
 const APP_ID = 'com.github.amezin.ddterm';
 const APP_WMCLASS = 'Com.github.amezin.ddterm';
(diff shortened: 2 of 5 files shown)


amezin:
Thank you for your effort.

Looks good on the first sight, but I need to test it manually a bit.

ddterm/org.gnome.Shell.Extensions.ddterm.service
```diff
@@ -0,0 +1,3 @@
+[D-BUS Service]
+Name=org.gnome.Shell.Extensions.ddterm
```
amezin, line 2: There's currently nothing with well-known name `org.gnome.Shell.Extensions.ddterm`. The app registers on the bus under the name `com.github.amezin.ddterm` https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/ddterm/app/application.js#L310

ddterm/shell/extension.js
```diff
@@ -245,8 +240,28 @@ function enable() {
...
+        Me.dir.get_child('ddterm').get_child('org.gnome.Shell.Extensions.ddterm.service'),
+        Gio.File.new_for_path(GLib.build_filenamev(
+            [
+                GLib.get_user_runtime_dir(),
+                'dbus-1',
+                'services',
+                'org.gnome.Shell.Extensions.ddterm.service',
+            ]))
+    );
+    dbus_service.install();
```
amezin, line 264: d-bus daemon docs for `$XDG_RUNTIME_DIR/dbus-1/services`:

> Also unlike the other standard session service directories, this directory is never monitored with inotify(7) or similar APIs. Programs that create service files in this directory while a dbus-daemon is running are expected to call the dbus-daemon's ReloadConfig() method after they have made changes.

https://dbus.freedesktop.org/doc/dbus-daemon.1.html

amezin (changes requested):
Try testing D-Bus activation using d-feet. When ddterm app isn't running (i. e. all tabs closed), d-feet should show the service on the session bus as activatable. And selecting it should start the application.

Also, when D-Bus activation starts working, it makes sense to enable it in the `.desktop` file.

Would `Exec=` in the `.desktop` file still be necessary?

pesader:
> Try testing D-Bus activation using d-feet. When ddterm app isn't running (i. e. all tabs closed), d-feet should show the service on the session bus as activatable. And selecting it should start the application.

Looks like it is working :D

https://user-images.githubusercontent.com/65264536/217408480-35611841-4725-488e-b48f-f3127710e2fb.mp4

> Also, when D-Bus activation starts working, it makes sense to enable it in the .desktop file.

Done in 5a7ea73.

> Would Exec= in the .desktop file still be necessary?

From what I read in the [Freedesktop Spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html), the Exec line must be kept to maintain compatibility with implementations that don't handle the DBusActivatable line properly. To quote them:

> If the value is true then implementations should ignore the Exec key and send a D-Bus message to launch the application. See [D-Bus Activation](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#dbus) for more information on how this works. Applications should still include Exec= lines in their desktop files for compatibility with implementations that do not understand the DBusActivatable key.

ddterm/shell/extension.js
pesader: Thanks, I did not know that! Done in 2f0acb9.

ddterm/org.gnome.Shell.Extensions.ddterm.service
pesader: In e0c2db5, I changed the well-known name (and the filename to match it).

amezin:
> Applications should still include Exec= lines in their desktop files for compatibility with implementations that do not understand the DBusActivatable key.

Hm. But the only "implementation" we need to care about is GNOME Shell.

pesader:
> Hm. But the only "implementation" we need to care about is GNOME Shell.

Yeah, makes sense. Your call then :)

PR #328, opened 2023-02-02
panelicon: unify toggle and menu panel icons

pesader:
Have a single option for panel icon, which toggles ddterm on middle or left click and displays the popup menu on right click. Works with both mouse and touchpad. Although this already "works", I wanted to know your opinion on a few additional changes that I believe could improve this PR.

**1. Have a single class for the panel icon**

To make it possible to have two options of panel icon behavior you wrote a `DDTermPanelIconBase` from which both `DDTermPanelIconPopupMenu` and `DDTermPanelIconToggleButton` inherit. In addition, there's a fourth class `DDTermPanelIconProxy` that instantiates the panel icon according to users' preferences.

If I understood the code correctly, unifying both panel icons should allow us to have a single `DDTermPanelIcon` class that takes care of all that.

**2. Use a toggle instead of multiselect in preferences**

Since there's only one option of panel icon, an on/off switch (boolean) would be more appropriate than the current multiselect (enum). I made a mockup of what it could look like if I also implemented your suggestion of making the icon itself configurable:

ddterm_preferences_icon_on
ddterm_preferences_icon_off

### DIFF
--- a/ddterm/pref/glade/prefs-panel-icon.ui
+++ b/ddterm/pref/glade/prefs-panel-icon.ui
@@ -54,7 +54,6 @@ along with ddterm GNOME Shell extension.  If not, see <http://www.gnu.org/licens
         <items>
           <item id="none" translatable="yes">None</item>
           <item id="menu-button" translatable="yes">With popup menu</item>
-          <item id="toggle-button" translatable="yes">Toggle button</item>
         </items>
       </object>
       <packing>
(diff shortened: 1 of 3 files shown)


amezin:
Could you add another mode instead of replacing all existing ones?

pesader:
Of course! That'll be much easier :D

amezin:
I mean, eventually, if it works well, other modes can be removed. But I'd prefer to have this change as a separate step (and maybe a separate release).

pesader:
I made another branch to add the new panel button, I'll open another PR :)

PR #349, opened 2023-03-06
panelicon: add option to have both toggle and popup menu

pesader:
As discussed in #328, I added a new panel button that includes the functionality of both previously available options (toggle and popup menu). I tested it using a touchpad and a mouse and it worked well on both cases.

I already ran the CI and it is failing at the step "Ensure translations are up to date", which I'm not sure how to fix 😅.

As usual, thanks for developing `ddterm`! 🚀

### DIFF
--- a/ddterm/pref/glade/prefs-panel-icon.ui
+++ b/ddterm/pref/glade/prefs-panel-icon.ui
@@ -55,6 +55,7 @@ along with ddterm GNOME Shell extension.  If not, see <http://www.gnu.org/licens
           <item id="none" translatable="yes">None</item>
           <item id="menu-button" translatable="yes">With popup menu</item>
           <item id="toggle-button" translatable="yes">Toggle button</item>
+          <item id="toggle-and-menu-button" translatable="yes">Toggle button and popup menu</item>
         </items>
       </object>
       <packing>
--- a/ddterm/shell/panelicon.js
+++ b/ddterm/shell/panelicon.js
@@ -145,6 +145,69 @@ const PanelIconToggleButton = GObject.registerClass(
     }
 );
 
+const PanelIconToggleAndMenu = GObject.registerClass(
+    class DDTermPanelIconToggleAndMenu extends PanelIconBase {
+        _init() {
+            super._init(false);
+
+            this.toggle_item = new PopupMenu.PopupSwitchMenuItem(
+                translations.gettext('Show'),
+                false
+            );
+            this.menu.addMenuItem(this.toggle_item);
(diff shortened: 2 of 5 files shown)


amezin:
To fix translations, you need to run `make msgmerge` and add the changes to your commit (empty translation stubs for UI labels you've added)

pesader:
I ran it, but I'm getting this error on the CI:

```log
Error: po/fr.po:819: this message needs to be reviewed by the translator
```

I checked out the file location and found this:

```po
#: ddterm/pref/glade/prefs-panel-icon.ui:57
msgid "Toggle button"
msgstr "Bouton de basculement"

#: ddterm/pref/glade/prefs-panel-icon.ui:58
#, fuzzy
msgid "Toggle button and popup menu"
msgstr "Bouton de basculement" # this line 819
```

Shouldn't that be empty?

amezin:
Yes, it should be empty.

Likely I should rethink how translations are maintained/processed.

For now you can remove `#, fuzzy` line, and the "automatically guessed" translation

pesader:
From reading the [docs](https://www.gnu.org/software/gettext/manual/html_node/msgmerge-Invocation.html), it looks like it is possible to disable fuzzy matching with the flag `--no-fuzzy-matching`. Maybe that would solve it :)

amezin:
Thank you and sorry for the inconvenience with translations

amezin:
Could you refactor `PanelIconToggleAndMenu` to inherit from `PanelIconPopupMenu` to get rid of duplicate code?

amezin:
> 1. There's also a lot of code from `PanelIconToggleButton`, but I don't feel like inheriting just from `PanelIconPopupMenu` makes that clear.

The menu has higher chance of being changed. Right now, if you want to add a menu item, you'll have to do it in two places. And there's not much that can be changed in a simple toggle button, so duplication here is a smaller problem.

> 2. We [discussed](https://github.com/ddterm/gnome-shell-extension-ddterm/pull/328#issuecomment-1426180789) merging both `PanelIconPopupMenu` and `PanelIconToggleButton` later on, as a separate change, and inhenritance would make that transition more difficult.

Not merging, just removing `PanelIconToggleButton` and configuration options. And `PanelIconPopupMenu` can remain as a base class, or can be merged back into `PanelIconToggleAndMenu`.

pesader:
> The menu has higher chance of being changed. Right now, if you want to add a menu item, you'll have to do it in two places.

Yeah makes sense. Thanks for explaining! I'll open another issue so we don't lose track of it.

PR #354, opened 2023-03-06
fr.po update

arnaud-feldmann:
I've put the requested translation and also, if you want, the translation of the short presentation for the app.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: ddterm\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-03-06 10:11-0300\n"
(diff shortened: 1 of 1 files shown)


arnaud-feldmann:
I have not sticked exactly to the "Another" cause it cannot be translated in french. The closest would be "Une extension parmi d'autres" but then it would be quite pejorative cause it would imply it's a forgettable extension.

amezin:
You're too far ahead with translating the `.desktop` file...

Right now it won't work, translations will need additional step to be integrated into `.desktop` file back.

Can I ask you to only translate the mentioned message for now?

`make po/ddterm@amezin.github.com.pot` should currently generate a .pot file without any mentions of `.desktop`

arnaud-feldmann:
@amezin I've suppressed the unwanted line, sticking to the .pot

Issue #87, opened 2021-10-01
Tabs get focus


lfom:
Hello!

Thanks for ddterm, I have been using since a while ago and I really enjoy it. Thanks a lot for this useful extension!

Small glitch/bug: clicking on the tabs with the mouse a second time focus the tab switch label, so what is typed does not go to the terminal. What is even weirder is that if you do it when a new tab is created then I can see the dashed line showing the focus, but if I hide then show again the terminal it behaves the same way but the dashed line is gone. I have played with the .ui files trying to disable the focus but it did not work, maybe it is hard coded somewhere?

Regards

amezin:
These are two separate issues. Could you report them as such? i. e. "New issue" and move one of the two paragraphs there.

lfom:
Done, search glitch moved to #89.

lfom:
Yes, GTK notebook is a little weird but I am sure this can be done. Both Gnome Terminal and Terminal use it for tabs and in both the tabs are not focuses and clicked, although you can still click (of couse) and double-click and secondary-click. Maybe setting the GTK property has_focus/can_focus in the code with prevent this?

amezin:
Fix will be included in the next release

PR #357, opened 2023-03-07
appwindow: make sure tabs are not focusable

pesader:
Fix issue #87.

### DIFF
--- a/ddterm/app/appwindow.js
+++ b/ddterm/app/appwindow.js
@@ -411,6 +411,7 @@ var AppWindow = GObject.registerClass(
             const index = this.notebook.insert_page(page, page.tab_label, position);
             this.notebook.set_current_page(index);
             this.notebook.set_tab_reorderable(page, true);
+            this.notebook.set_can_focus(false);
 
             const page_scope = rxutil.scope(page, rxutil.signal(page, 'destroy'));
             this.rx.add(page_scope);


amezin:
Looking at GNOME Terminal source code:

It [overrides](https://github.com/GNOME/gnome-terminal/blob/70a484db960e4ef078ea1d8c2c96a1c0cbfac1e2/src/terminal-notebook.cc#L488) notebook widget's [`grab_focus` virtual method](https://docs.gtk.org/gtk3/vfunc.Widget.grab_focus.html), and [redirects it to the active terminal widget](https://github.com/GNOME/gnome-terminal/blob/master/src/terminal-notebook.cc#L375-L382). And does not touch `can-focus` property.

I wonder if there is any practical difference (usually for UI I try to replicate GNOME Terminal - unless there are reasons not to do so).

pesader:
Hey, sorry for disappearing! Classes are already at full speed, so I haven't had the time to make the changes you requested.

I saw you merged my PR already though. Will we keep it that way, or do you still prefer to reimplement the fix using GNOME Terminal's method?

amezin:
> I saw you merged my PR already though. Will we keep it that way, or do you still prefer to reimplement the fix using GNOME Terminal's method?

Currently, most of the UI is loaded from Glade (`.ui`) files. Overriding `grab_focus` will be a non-trivial change (even though moving from `.ui` files to just code seems to be the right direction). So I decided to accept it in this form for now.

PR #359, opened 2023-03-08
translations: disable msgmerge fuzzy matching

pesader:
Create the variable `MSGMERGE_FLAGS` and add the flag `--no-fuzzy-matching` to it, so only exact matches are used when updating the .po files. Also move the `-U` flag to the same variable, using its non-abbreviated form `--update` for improved clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -339,9 +339,10 @@ msgcmp-strict: $(MSGCMP_GOALS)
 all: msgcmp
 
 MSGMERGE_GOALS := $(addprefix msgmerge/, $(LOCALES))
+MSGMERGE_FLAGS := --no-fuzzy-matching --update
 
 $(MSGMERGE_GOALS): msgmerge/%: % $(POT_FILE)
-	msgmerge -U $^
+	msgmerge $(MSGMERGE_FLAGS) $^
 
 msgmerge: $(MSGMERGE_GOALS)
 
--- a/Makefile
+++ b/Makefile
@@ -346,6 +346,9 @@ $(MSGMERGE_GOALS): msgmerge/%: % $(POT_FILE)
 
 msgmerge: $(MSGMERGE_GOALS)
 
+msgmerge-fuzzy: MSGMERGE_FLAGS := --update
+msgmerge-fuzzy: $(MSGMERGE_GOALS)
+
 .PHONY: msgmerge $(MSGMERGE_GOALS)
 
 # ESLint


amezin:
Look at msgcmp target: https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/Makefile#L332-L336

Because sometimes you want the flag `--use-untranslated`, and sometimes you don't, I've made two targets. Could you do the same for `msgmerge` (for example, `msgmerge-fuzzy`)? Because sometimes fuzzy matching may be desirable.

pesader:
Thanks for the feedback. To be honest, I'm not very well versed in the syntax of Makefiles.
Did you mean something like this?

```makefile
MSGMERGE_GOALS := $(addprefix msgmerge/, $(LOCALES))

msgmerge: MSGMERGE_FLAGS := --no-fuzzy-matching --update
msgmerge: $(MSGMERGE_GOALS)

msgmerge-fuzzy: MSGMERGE_FLAGS := --update
msgmerge-fuzzy: $(MSGMERGE_GOALS)
```

amezin:
Yes, something like that.

Although, I've changed msgcmp a bit: https://github.com/ddterm/gnome-shell-extension-ddterm/commit/131a50ba63ab04dd1dd77a40b0ac4f918321ae3a

Turns out command line has precedence over target-specific variables too, so half of the commit message is incorrect (yep, I'm not a makefile expert too). But the second half is still correct.

And all of this may be more or less pointless soon because I'm trying to move all translation work to weblate https://hosted.weblate.org/projects/gnome-shell-extension-ddterm/ (it has a built-in `msgmerge` hook https://github.com/ddterm/gnome-shell-extension-ddterm/pull/363/commits/e03e85514465492034c4aedac9def2a2b2adc24b)

pesader:
> yep, I'm not a makefile expert too

Doesn't look like it! `ddterm`'s Makefile is what I use as a reference for my own projects :)

> And all of this may be more or less pointless soon because I'm trying to move all translation work to weblate

That's great news. I organized some FOSS translation workshops in my university and Weblate makes it all the more accessible for people who aren't used to the Git/GitHub contribution workflow. Since it seems this minor change won't be necessary anymore, I'll close the PR.

Let me know when the Weblate infrastructure is ready. I'll translate `ddterm` to Brazilian Portuguese, so my peers can use it in their mother language 😉 🇧🇷

amezin:
Not completely "ready", but it's already working, and somebody already started Norwegian Bokmål translation.

> Since it seems this minor change won't be necessary anymore, I'll close the PR.

I'm not removing `msgcmp` targets yet, and it would be better if `msgmerge` was consistent with `msgcmp`. If you have the time, I'd appreciate if you finished this change.

pesader:
> If you have the time, I'd appreciate if you finished this change.

Cool :)
Let me know if this is what you were looking for.

PR #422, opened 2023-04-21
fix: translate filename (Chinese (Simplified))

yuhldr:


amezin:
Hm, what's going on? Is Weblate generating wrong file names?

yuhldr:
> Hm, what's going on? Is Weblate generating wrong file names?

yes，like this [GNOME-console -po/zh_CN.po](https://gitlab.gnome.org/GNOME/console/-/blob/main/po/zh_CN.po)

amezin:
I think I've fixed Weblate configuration, but, unfortunately, the fix will only apply to new translations. For the existing one we'll have to do it manually

PR #498, opened 2023-07-22
fix do-in-docker args

Zerocool56:
Add :z flag to volumes. It's fix permission problems on hosts with SELinux (like Fedora 38)
Add HOME env. It's fix permission problems with creation ~/.npm dir

### DIFF
--- a/do-in-docker.sh
+++ b/do-in-docker.sh
@@ -17,4 +17,4 @@ fi
 
 set -ex
 
-exec "${docker_cmd[@]}" run --init --rm -i $TTY_FLAG -u $UID_GID -e HOME=${HOME_DIR} -v "${SCRIPT_DIR}:${SCRIPT_DIR}:z" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
+exec "${docker_cmd[@]}" run --init --rm -i $TTY_FLAG -u $UID_GID -e HOME="${HOME_DIR}" --security-opt label=disable -v "${SCRIPT_DIR}:${SCRIPT_DIR}" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
--- a/do-in-docker.sh
+++ b/do-in-docker.sh
@@ -17,4 +17,4 @@ fi
 
 set -ex
 
-exec "${docker_cmd[@]}" run --init --rm -i $TTY_FLAG -u $UID_GID -v "${SCRIPT_DIR}:${SCRIPT_DIR}" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
+exec "${docker_cmd[@]}" run --init --rm -i $TTY_FLAG -u $UID_GID -e HOME=${PWD} -v "${SCRIPT_DIR}:${SCRIPT_DIR}:z" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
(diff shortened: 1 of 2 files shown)


amezin:
So it happens only with `do-in-docker.sh` because CI runs everything as root, and podman maps host uid to container's root.
But `do-in-docker.sh` keeps host uid which has no corresponding user in the container (no home, no passwd entry, etc).

Storing `HOME` content directly in the working directory doesn't seem like a good idea - at least because that content won't be ignored by git, eslint, etc.

Unless there's a specific reason to keep the content between do-in-docker.sh invocations, I'd prefer `HOME` to be a temporary directory (it will match `do-in-podman.sh` behavior). Otherwise, it should be all kept in a subdirectory (for example, `${SCRIPT_DIR}/.container-home`), and that directory should be added to `.gitignore`. In this case, `do-in-podman.sh` should also be modified accordingly.

Also, could you explain SELinux issue in more detail?

Zerocool56:
About selinux
https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label
As said in official docs, if we mount dir into the container, we should use z option to update selinux labels for correct permissions.
On my fedora38 without this flag i've got a permission error on npm install step.

amezin:
I don't like the idea of changing permissions (or other attributes) of the files. At least when it could be avoided.

With podman, `--security-opt label=disable` works without `:z`. Since it's not some public production service accessed from an untrusted network, I think `--security-opt label=disable` should be fine.

I can't reproduce the issue with Docker though. Installed as in https://developer.fedoraproject.org/tools/docker/docker-installation.html Have you enabled SELinux support in Docker manually?

do-in-docker.sh
```diff
@@ -17,4 +17,4 @@ fi
 
 set -ex
 
-exec "${docker_cmd[@]}" run --init --rm -i $TTY_FLAG -u $UID_GID -v "${SCRIPT_DIR}:${SCRIPT_DIR}" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
+exec "${docker_cmd[@]}" run --init --rm -i $TTY_FLAG -u $UID_GID -e HOME=${HOME_DIR} -v "${SCRIPT_DIR}:${SCRIPT_DIR}:z" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
```
amezin, line 20: Please add quotes around `HOME=${HOME_DIR}`. It could contain spaces.

do-in-podman.sh
```diff
@@ -5,9 +5,10 @@ IMAGE=ghcr.io/ddterm/ci-docker-image
...
 
 SCRIPT_DIR=$(CDPATH="" cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+HOME_DIR="${SCRIPT_DIR}/.container-home"
 
 TTY_FLAG=$(test -t 0 && echo -n -t)
 
 set -ex
 
-exec podman run --init --rm -i $TTY_FLAG -v "${SCRIPT_DIR}:${SCRIPT_DIR}" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
+exec podman run --init --rm -i $TTY_FLAG -e HOME=${HOME_DIR} -v "${SCRIPT_DIR}:${SCRIPT_DIR}:z" -w "${PWD}" "${IMAGE}:${IMAGE_VERSION}" xvfb-run "$@"
```
amezin, line 14: Please add quotes around `HOME=${HOME_DIR}`. It could contain spaces.

Instead of relabeling, please do `--security-opt label=disable`.

amezin (changes requested):

Zerocool56:
I just install docker from the base repos of Fedora, simply run `dnf install docker` but I suppose it's not necessary.
And SELinux was enabled by default.

About security opt, maybe it's more secure way and we can use it in both situations

amezin:
> About security opt, maybe it's more secure way

It's "less secure", effectively disabling SELinux. Not disabling completely, just not limiting permissions additionally.

Zerocool56:
I meant more secure than relabeling dirs and files.

PR #502, opened 2023-07-24
Makefile: Expand wildcard

bunder:
Without this patch build fails with following error:

```
make: *** No rule to make target 'ddterm/app/icons/*', needed by '.../ddterm@amezin.github.com/ddterm/app/icons/*'.  Stop.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,7 @@ PACK_CONTENT += $(EXECUTABLES) $(filter-out $(EXECUTABLES),$(JS_SOURCES))
 PACK_CONTENT += \
 	ddterm/app/style.css \
 	ddterm/app/dependencies.json \
-	ddterm/app/icons/* \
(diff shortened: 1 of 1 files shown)


amezin:
Weird. It builds successfully on my machine and in CI.

I can trigger this error only when `ddterm/app/icons/` is empty. But it shouldn't be empty.

bunder:
`ddterm/app/icons` is not empty on my machine. I have somewhat standard setup: Arch Linux, x86_64, bash, GNU Make 4.4.1.

amezin:
> But you're right: make should expand wildcards automatically, when they are used in rules. I don't know why this doesn't work.

At least for system-wide "make install", it happens because

```
SYS_INSTALLED_CONTENT := $(addprefix $(SYS_INSTALLED_FULL_PREFIX)/,$(PACK_CONTENT))
```

Wildcards won't work for "destination" files obviously

PR #566, opened 2023-09-26
add Greek translation

rizitis:


amezin:
Could you upload it to Weblate instead (you should be able to upload the same .po file there)?

For me it'll be a bit more convenient, because I'll be able to see how many strings are translated (I see that at least some strings aren't translated so far).

Weblate will generate a commit with your authorship

rizitis:
I m not familiar with Weblate, I just created a Greek translation there but I cant find a way to upload el.po, only to start translating from scratch, which I ll not...  If there is a way el.po to be uploaded there plz do it...

amezin:
https://hosted.weblate.org/projects/gnome-shell-extension-ddterm/extension/el/#upload doesn't this work?

rizitis:
probably yes, can you confirm?

amezin:
Ok, I've uploaded the file myself

This resulted in https://github.com/ddterm/gnome-shell-extension-ddterm/pull/479/commits/2a8ad0c28f30218a666e9a182b4c39baab5f545f

Is https://github.com/imacforlinux your other account?

Other than that, the translation seems to be almost complete, I'm ok with merging it

amezin:
hm

ok, let's merge this PR and see how Weblate handles that

It's weird that it added someone (unrelated?)

PR #745, opened 2024-01-12
extension: expose Hide action in D-Bus API

SyntaxBlitz:
I have some scripts that I mostly only run from the dropdown terminal, and I like them to be able to close the dropdown automatically when they finish. `Toggle` works, but on the rare occasions when I run these scripts from other terminals, I don't want it to _open_ the dropdown. Using `Activate` then `Toggle` is a workaround, but it causes the terminal to flicker when the dropdown is not already active.

This PR exposes the idempotent `Hide` action over the D-Bus API. I think I hit all the spots where the code needed to be updated -- I just grepped `toggle` and `activate` and looked for all the spots where it looked like I needed to wire things up. Apologies if I missed anything; I haven't fully wrapped my head around the extension architecture, so this is mostly a copy-paste job.

I wrote this against the v46 tag because I'm running GNOME shell 42.9 (on `jammy`), but it merges fine against master as well. Tested manually on my machine with:

```
gdbus call --session --dest=org.gnome.Shell --object-path /org/gnome/Shell/Extensions/ddterm --method com.github.amezin.ddterm.Extension.Hide
```

which seems to work the same way as `.Activate`. I don't see anything breaking if I call it when the terminal is already hidden.

### DIFF
--- a/ddterm/com.github.amezin.ddterm.Extension.xml
+++ b/ddterm/com.github.amezin.ddterm.Extension.xml
@@ -5,6 +5,7 @@
     <interface name="com.github.amezin.ddterm.Extension">
         <method name="Toggle"/>
         <method name="Activate"/>
+        <method name="Hide"/>
         <method name="Service"/>
         <method name="GetTargetRect">
             <arg type="i" name="x" direction="out"/>
--- a/ddterm/shell/dbusapi.js
+++ b/ddterm/shell/dbusapi.js
@@ -92,6 +92,10 @@ var Api = GObject.registerClass({
             return_type: GObject.TYPE_JSOBJECT,
             accumulator: GObject.AccumulatorType.FIRST_WINS,
         },
+        'hide': {
+            return_type: GObject.TYPE_JSOBJECT,
+            accumulator: GObject.AccumulatorType.FIRST_WINS,
+        },
         'service': {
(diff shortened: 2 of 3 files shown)


amezin:
Does [`gapplication action com.github.amezin.ddterm hide`](https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/docs/cmdline.md#gapplication) not work for you?

Doing this through the extension API may have a few positive effects - like not starting the app if it's not running, and waiting for the window to actually hide. But is it necessary in your case?

By the way, v47 should also work fine on GNOME 42, if I'm not mistaken.

SyntaxBlitz:
Ah, thanks for the pointer! I jumped to D-Bus because the last extension I used (before I upgraded to `jammy` yesterday) only had a D-Bus API; when I saw this extension also had D-Bus listeners, that's where I started looking for a `hide` action. Indeed, the extension API you mentioned does work for me.

v46 was a bit of an arbitrary choice I made from skimming the release notes, since I was more sure it would work. I figured this diff probably wouldn't touch the ESM updates either way.

Wonderful extension! Thanks again.

amezin:
On the other hand, adding this method shouldn't cause any issues, and it'll make the interface logically complete.

PR #808, opened 2024-02-25
Enable GNOME 46 support

amezin:
Fixes https://github.com/ddterm/gnome-shell-extension-ddterm/issues/807


amezin:
46.rc breaks notification api: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3173

PR #889, opened 2024-04-15
Allow configuring window size (width/height)

martinn:
First of all, thanks for a great extension!

This PR adds the ability to configure both horizontal and vertical window size (previously, only one was configurable - depending on your window orientation). It can be very useful for people with large/ultrawide monitors.

For now, the window position is always centered, but that could be made configurable in a future iteration, if desired.

Screenshot from 2024-04-15 12-17-07
Screenshot from 2024-04-15 12-21-06
Screenshot from 2024-04-15 12-21-36
Screenshot from 2024-04-15 12-22-21

I still need to dive into the automated testing setup and add/update tests for this change, but wanted to get some thoughts/feedback to see if happy with the general approach before I do that.

I don't really know a whole lot about gnome shell extension development so please just let me know if any issues at all.

Thanks!

Resolves: https://github.com/ddterm/gnome-shell-extension-ddterm/issues/129

### DIFF
--- a/test/test_extension.py
+++ b/test/test_extension.py
@@ -1,5 +1,6 @@
 import collections
 import contextlib
+import dataclasses
 import enum
 import functools
 import logging.handlers
@@ -90,21 +91,21 @@ def resize_point(frame_rect, window_pos):
 
 def compute_target_rect(size, pos, monitor):
     x, y, width, height = monitor.workarea
+    window_size = WindowSize.get_maximized_size_from_position(size, pos)
 
     round_to = int(monitor.scale)
 
-    if pos in [WindowPosition.TOP, WindowPosition.BOTTOM]:
-        height *= size
-        height -= height % round_to
+    height *= window_size.height
+    height -= height % round_to
 
-        if pos == WindowPosition.BOTTOM:
-            y += monitor.workarea.height - height
-    else:
-        width *= size
-        width -= width % round_to
+    if pos == WindowPosition.BOTTOM:
+        y += monitor.workarea.height - height
 
-        if pos == WindowPosition.RIGHT:
-            x += monitor.workarea.width - width
+    width *= window_size.width
+    width -= width % round_to
+
+    if pos == WindowPosition.RIGHT:
+        x += monitor.workarea.width - width
 
     return Rect(x, y, width, height)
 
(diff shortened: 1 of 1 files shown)


amezin:
Does this work when both width and height are set to 90%?

amezin:
> It should. I just tried it and seems to work fine on my end, though I do get a second resizing happening for some reason (i.e. first it tries to animate to almost 100% size, then it resizes back to 90%). I also see some odd behavior where if you set size to 100%, it ends up resizing to less than that. I'll look into it further.

That's what stopped me from implementing this feature previously, because I was unable to find a working solution for this problem.

amezin:
> Do you mean the first part (issues at 90%), the second part (issues at 100%) or both?

Honestly, I can't recall all the details. But it was all around (un)maximization.

> I also saw some parts of the codebase that talk about maximizing the window in some aspect. If you've got time, are you able to share a bit more about how that works?

First of all, GNOME's window manager - mutter - has its own internal "smart" logic about maximization, about window placement, etc. For example, if a window covers >=80% of the screen (excluding the panel, so-called "workarea"), the window manager maximizes the window automatically. Also, it sometimes moves the window - either to a different location on the same monitor, or even to a different monitor.  So the extension tries to cancel all that - unmaximizes the window if it shouldn't be maximized, tracks the window geometry and "fixes" it if/when necessary. To make things worse, mutter has some issues with synchronization on Wayland https://gitlab.gnome.org/GNOME/mutter/-/issues/1627 - so it ignores some move/resize requests when it shouldn't

amezin:
It would be a bit easier if you kept `window-size` setting as is, i.e. to always be the "primary size". Because you wouldn't need to modify tests in this case.

Maybe disabling [`Meta.Preference.AUTO_MAXIMIZE`](https://gjs-docs.gnome.org/meta12~12/meta.preference#default-auto_maximize) temporarily could help fighting the unnecessary maximization.

But, honestly, I doubt this entire feature is feasible with current state of Mutter.

amezin:
BTW, if you have any idea how to make test code less awful, even a bit, please let me know

amezin:
> I guess I could move this logic to the extension

Which logic exactly? The extension already chooses to apply `window-size` to either width or height depending on the position.

> Swap the window width/height values if changing from top/bottom to left/right and viceversa. Unfortunately the UI does not show the updated values and I'm not sure how to make it refresh (any tips here?)

Because you're not changing the settings. All `WindowGeometry` settings bindings are "read-only": https://github.com/ddterm/gnome-shell-extension-ddterm/blob/02c962ed213c93976db0d276ef32268c22581e98/ddterm/shell/geometry.js#L167

And I'd prefer them to stay read-only. This class was meant to transform settings to window geometry only in one direction.

amezin:
@martinn Sorry, have you given up? (To be honest, I tried to solve this issue once, and given up)

BTW, if tests are failing, but in real use things are working as expected - that might be a mistake in the tests.

JE-Amouel:
any update about this feature?

PR #981, opened 2024-07-03
Fix warning on GLib >=2.79.2 

C-512L:


amezin:
Your change isn't compatible with GNOME versions older than 46. ddterm `master` branch currently supports GNOME >= 43. Between supporting more GNOME versions, and getting rid of a warning, I will obviously choose compatibility. 

But you likely can make it work for all GNOME versions using https://gjs-docs.gnome.org/gjs/overrides.md#gi-require

C-512L:
I have been able to backport it to all systems using ESM on my own branch, but I might have  stumbled on a ESM translator edge case being unable to translate `import Gi from 'gi';` correctly. I tried to look at that code of `translate-esm.js` but unfortunalety I haven't found much documentation about using `gi.require()` on pre-ESM GJS to make patch myself.

amezin:
There's also sed preprocessing step to solve issues like that: https://github.com/ddterm/gnome-shell-extension-ddterm/blob/bfcdf692fe2e1e907556aa6243ebc5d50de3fa31/ddterm/shell/notifications.js#L143-L151

You should be able to use legacy imports between `// BEGIN !ESM` and `// END !ESM` lines

C-512L:
I didn't feel sure about adding a lint exception, but I guessed it would be a better solution than adding dead code to the non-ESM build or duplicating code. Feedback is welcomed.

ddterm/shell/sd_journal.js
```diff
@@ -75,7 +79,17 @@ export function sd_journal_stream_fd(identifier, priority = LOG_INFO, level_pref
     }
 
     try {
-        GLib.unix_set_fd_nonblocking(fd, false);
+        // BEGIN ESM
+        if (major > 45) {
```
amezin, line 83: Maybe just wrap `Gi.require()` in `try {} catch`? It seems that the exception thrown by `Gi.require()` isn't `GLib.Error`, so it should be easy to distinguish from `set_fd_nonblocking()` failure.

ddterm/shell/sd_journal.js
```diff
@@ -75,7 +79,17 @@ export function sd_journal_stream_fd(identifier, priority = LOG_INFO, level_pref
...
+            const GLibUnix = Gi.require('GLibUnix', '2.0');
+            GLibUnix.set_fd_nonblocking(fd, false);
+        } else {
+        // END ESM
+        // eslint-disable-next-line indent
+            GLib.unix_set_fd_nonblocking(fd, false);
```
amezin, line 89: Maybe `return fd` earlier and avoid `else {}`?

amezin (changes requested):

ddterm/shell/sd_journal.js
C-512L: I'm not sure I am a fan of having multiple nested `try {} catch` re-throwing errors. I think it adds complexity compared to the shell version check, as I don't see a real reason to make it dynamic if there is no need for more error handling.

ddterm/shell/sd_journal.js
C-512L: Could you elaborate a little further?

ddterm/shell/sd_journal.js
amezin: If you insist on version check, you should check the version of GLib, not Shell. https://github.com/GNOME/glib/commit/5021d002abe55e4d09e7e83b3ffe3a8cf1a074ac

ddterm/shell/sd_journal.js
amezin:
```suggestion
            GLibUnix.set_fd_nonblocking(fd, false);
            return fd;
        }
        // END ESM

        GLib.unix_set_fd_nonblocking(fd, false);
        return fd;
```

C-512L: Done :+1:

ddterm/shell/sd_journal.js
C-512L: > If you insist on version check, you should check the version of GLib, not Shell. [GNOME/glib@5021d00](https://github.com/GNOME/glib/commit/5021d002abe55e4d09e7e83b3ffe3a8cf1a074ac)

Done in my last push. :+1:

ddterm/shell/sd_journal.js
```diff
@@ -19,7 +19,9 @@
 
 import GLib from 'gi://GLib';
 import Gio from 'gi://Gio';
-
+// BEGIN ESM
+import Gi from 'gi';
+// END ESM
```
amezin, line 24:
```suggestion
// BEGIN ESM

import Gi from 'gi';
// END ESM

```

Different types of imports and other things should be separated by an empty line in both ESM and legacy packages

ddterm/shell/sd_journal.js
```diff
@@ -75,6 +77,13 @@ export function sd_journal_stream_fd(identifier, priority = LOG_INFO, level_pref
     }
 
     try {
+        // BEGIN ESM
+        if (GLib.check_version(2, 79, 2) === null) {
+            const GLibUnix = Gi.require('GLibUnix', '2.0');
+            GLibUnix.set_fd_nonblocking(fd, false);
+            return fd;
+        }
+        // END ESM
```
amezin, line 86:
```suggestion

        // END ESM
```

amezin (changes requested):

ddterm/shell/sd_journal.js
C-512L: Done. I hope I haven't missed any separation. 😅

PR #1057, opened 2024-09-29
app/window: fix target monitor detection

amezin:


amezin:
This doesn't really work correctly on older GNOME versions. https://github.com/ddterm/gnome-shell-extension-ddterm/pull/1058 seems to give much better results

PR #1132, opened 2024-12-13
fix: remove blank spacing from panel icons

yuhldr:
version|pic
-|-
old|截图 2024-12-13 19-48-01
new|截图 2024-12-13 19-47-13

### DIFF
--- a/ddterm/shell/panelicon.js
+++ b/ddterm/shell/panelicon.js
@@ -47,7 +47,7 @@ const PanelIconBase = GObject.registerClass({
 
         this.add_child(new St.Icon({
             icon_name: 'utilities-terminal',
-            style_class: 'system-status-icon',
+            style_class: 'system-status-icon indicator-icon',
         }));
     }
 });
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -1,3 +1,8 @@
 #ddterm-panel-icon > StIcon {
     -st-icon-style: symbolic;
 }
+
+.indicator-icon {
+  padding: 0px;
+  margin: 0px;
+}


ddterm/shell/panelicon.js
```diff
@@ -47,7 +47,7 @@ const PanelIconBase = GObject.registerClass({
 
         this.add_child(new St.Icon({
             icon_name: 'utilities-terminal',
-            style_class: 'system-status-icon',
+            style_class: 'system-status-icon indicator-icon',
```
amezin, line 50: Isn't there a built-in (Shell) style that does this? When other extensions add icons to the panel, is spacing similar to ddterm?

stylesheet.css
```diff
@@ -1,3 +1,8 @@
 #ddterm-panel-icon > StIcon {
     -st-icon-style: symbolic;
 }
+
+.indicator-icon {
```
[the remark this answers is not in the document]
yuhldr: like this, and test ok:

https://github.com/oae/gnome-shell-pano/blob/d0d71f8015ada661f35a1390502c8e09fa2a10b6/src/styles/stylesheet.css#L189-L192

ddterm/shell/panelicon.js
yuhldr: here:

https://github.com/oae/gnome-shell-pano/blob/d0d71f8015ada661f35a1390502c8e09fa2a10b6/src/components/indicator/settingsMenu.ts#L62

yuhldr: If we don't modify ddterm, using [Just Perfection](https://extensions.gnome.org/extension/3843/just-perfection/) will not achieve the desired effect and will result in inconsistent spacing between icons

stylesheet.css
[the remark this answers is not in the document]
yuhldr: This will only affect ddterm and will not affect other plugins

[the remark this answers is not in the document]
amezin: No, this will affect any UI element with `indicator-icon` style.

For example, if you add this to the stylesheet:

```css
.system-status-icon {
    icon-size: 8px;
}
```

all icons will become small, not only ddterm's one

ddterm/shell/panelicon.js
amezin: > If we don't modify ddterm, using [Just Perfection](https://extensions.gnome.org/extension/3843/just-perfection/) will not achieve the desired effect

What is "the desired effect"?

> will result in inconsistent spacing between icons

Current implementation is consistent with "Removable Drive Menu" from gnome-shell-extensions package: https://github.com/GNOME/gnome-shell-extensions/blob/4a841dfd4977c3d497b0241c7d0016c4290b3e07/extensions/drive-menu/extension.js#L144-L147

yuhldr: I test again, you are right

PR #1147, opened 2024-12-30
Save and restore terminal content

amezin:


amezin:
Needs manual testing on GNOME 42-43 and 44-45

PR #1148, opened 2024-12-30
tests: add Alpine 3.18 (GNOME 44)

amezin:


amezin:
Either need to figure out what's wrong with v44, or remove from supported versions before release

PR #1197, opened 2025-02-15
Add option to keep previous session as a backup

samuelba:
When enabled, this helps to prevent loosing the session in case of an unexpected OS/app crash. I have many pre-configured tabs, that I already lost many times and I always have to set them up again, which is cumbersome.

I added it as an option to not modify the current (intentional?) behavior. Of course we could also just simply remove the line `GLib.unlink(this.session_file_path);`.

### DIFF
--- a/ddterm/app/application.js
+++ b/ddterm/app/application.js
@@ -662,7 +662,8 @@ class Application extends Gtk.Application {
             this.ensure_window().deserialize_state(data_variant);
         }
 
-        GLib.unlink(this.session_file_path);
+        if (!this.settings.get_boolean('keep-session-backup'))
+            GLib.unlink(this.session_file_path);
     }
 
     save_session() {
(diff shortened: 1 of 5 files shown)


amezin:
Hm.

Maybe there's actually no need to `.unlink()`.

Or it should be done only if:

- saving the session with no tabs (this happens when you close all tabs manually)

- there's an error when restoring the session. I think `.unlink()` was initially added as a protection against broken session files, but this can be done with `try { } catch { }` instead.

I definitely don't like the idea of adding another setting/option, the "default" (and only) behavior should be better here.

amezin:
Also, you may save the session in more places (for example, every time when a tab is added or removed). But it should be a separate PR, after we figure out what to do with `.unlink()`.

samuelba:
I would also be in favor of removing `unlink()`, rather than having the option. I can adapt it to fulfill your two points.

> Also, you may save the session in more places (for example, every time when a tab is added or removed). But it should be a separate PR, after we figure out what to do with .unlink().

Good points and I agree that would be a separate MR.

samuelba:
close in favor of https://github.com/ddterm/gnome-shell-extension-ddterm/pull/1198

PR #1198, opened 2025-02-15
app: don't remove session file after loading it

samuelba:
Do not delete the session file after successfully restoring it. This has the benefit that we do not loose the session in case of a OS/app crash.

Delete the session file if all tabs are closed (windows destroyed).

Delete the session file in case something goes wrong during restoring the session.

I actually could not make the session restore to fail, if I modified a session file it would just not load it and start with a new tab. But it did not throw  any errors or so, which would delete the file. But I guess it cannot hurt to have that in place anyway.

### DIFF
--- a/ddterm/app/application.js
+++ b/ddterm/app/application.js
@@ -350,8 +350,6 @@ class Application extends Gtk.Application {
             if (!(ex instanceof GLib.Error &&
                 ex.matches(GLib.file_error_quark(), GLib.FileError.NOENT)))
                 logError(ex, "Can't restore session");
-
-            GLib.unlink(this.session_file_path);
         }
 
         this.connect('query-end', () => {
@@ -554,12 +552,13 @@ class Application extends Gtk.Application {
             display_config: this.display_config,
         });
 
+        this.window.connect('close', () => {
+            this.save_session();
+        });
+
         this.window.connect('destroy', source => {
-            if (source === this.window) {
-                // All tabs are closed, remove the session file
-                GLib.unlink(this.session_file_path);
+            if (source === this.window)
                 this.window = null;
-            }
         });
 
         return this.window;
@@ -655,9 +654,7 @@ class Application extends Gtk.Application {
     }
 
     restore_session() {
-        const [ok, data] = GLib.file_get_contents(this.session_file_path);
-        if (!ok)
-            GLib.unlink(this.session_file_path);
+        const [, data] = GLib.file_get_contents(this.session_file_path);
 
         if (data?.length) {
             const data_variant = GLib.Variant.new_from_bytes(
@@ -672,6 +669,11 @@ class Application extends Gtk.Application {
 
     save_session() {
         const data = this.window?.serialize_state();
+        if (!data) {
+            GLib.unlink(this.session_file_path);
+            return;
+        }
+
         const bytes = data?.get_data_as_bytes().toArray() ?? [];
 
         GLib.mkdir_with_parents(GLib.path_get_dirname(this.session_file_path), 0o700);
--- a/ddterm/app/application.js
+++ b/ddterm/app/application.js
@@ -655,14 +655,14 @@ class Application extends Gtk.Application {
                 );
 
                 if (!data_variant.is_normal_form())
-                    throw new Error('Invalid session file data variant, delete session file.');
+                    throw new Error('Session data is malformed, probably the file was damaged');
 
                 this.ensure_window().deserialize_state(data_variant);
             }
         } catch (ex) {
             if (!(ex instanceof GLib.Error &&
                 ex.matches(GLib.file_error_quark(), GLib.FileError.NOENT))) {
-                logError(ex, "Can't restore session, delete session file.");
+                logError(ex, "Can't restore session. Deleting session file.");
                 GLib.unlink(this.session_file_path);
             }
         }
--- a/ddterm/app/application.js
+++ b/ddterm/app/application.js
@@ -350,6 +350,8 @@ class Application extends Gtk.Application {
             if (!(ex instanceof GLib.Error &&
                 ex.matches(GLib.file_error_quark(), GLib.FileError.NOENT)))
                 logError(ex, "Can't restore session");
+
+            GLib.unlink(this.session_file_path);
         }
 
         this.connect('query-end', () => {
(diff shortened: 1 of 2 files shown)


amezin:
I'll need some time to review, the logic is non-trivial. To make things worse, there are no tests that verify this code right now.

Could you add some tests, what do you think? Absolutely not required for the PR to be accepted, because the test code is a mess currently (and likely will always be) - but it would help a lot

samuelba:
Some unit tests would be nice, but I don't think I manage to add one just yet from scratch without another application.js unit test that I could use for reference.

amezin:
There are app tests, but they can't really be called "unit tests". More like functional or end-to-end tests. They launch a complete GNOME Shell session and simulate mouse clicks, perform CLI and D-Bus calls, etc. And they are written in Python not JS https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/tests/test_app.py

ddterm/app/application.js
```diff
@@ -553,8 +555,11 @@ class Application extends Gtk.Application {
         });
 
         this.window.connect('destroy', source => {
-            if (source === this.window)
+            if (source === this.window) {
+                // All tabs are closed, remove the session file
+                GLib.unlink(this.session_file_path);
```
amezin, line 560: Doesn't it remove the session file every time the window is closed? Including, for example, when you log out while ddterm is running (so the session gets saved and then immediately removed).

IMO the file should be removed in `save_session()` instead - when there's nothing to save, instead of writing an empty file.

ddterm/app/application.js
```diff
@@ -650,7 +655,9 @@ class Application extends Gtk.Application {
     }
 
     restore_session() {
-        const [, data] = GLib.file_get_contents(this.session_file_path);
+        const [ok, data] = GLib.file_get_contents(this.session_file_path);
+        if (!ok)
+            GLib.unlink(this.session_file_path);
```
amezin, line 660: This will never happen. `ok` will be always `true` here. Instead, `file_get_contents()` will `throw` if there's an I/O error. Also, I'm not sure if it's a good idea to remove the file on I/O error.

Instead, the file should be removed if its content is invalid - i.e. if it can't be parsed as `GLib.Variant` or if something goes wrong when you try to restore the session from it.

ddterm/app/application.js
```diff
@@ -350,6 +350,8 @@ class Application extends Gtk.Application {
             if (!(ex instanceof GLib.Error &&
                 ex.matches(GLib.file_error_quark(), GLib.FileError.NOENT)))
                 logError(ex, "Can't restore session");
+
+            GLib.unlink(this.session_file_path);
```
amezin, line 354: This line will be executed even if the error is `GLib.FileError.NOENT`. Or any other I/O error. I don't think removing the file here is a good idea.

amezin (changes requested):

ddterm/app/application.js
amezin: Or maybe forget about this case (no tabs/empty file) for now. Existing code writes an empty file, and it doesn't cause any problems. This can be a separate "optimization" PR.

Just don't remove the session file on `destroy`.

amezin: Tested this branch again and, to my surprise, the session gets saved. Apparently, `destroy` isn't emitted when the session manager terminates the app. But when closing all tabs, an empty file gets written from `save_session()` first, then it's removed here - not the best solution.

amezin: Oh, `.save_session()` actually doesn't get called when you close all tabs... You're right, `unlink()` is necessary when closing the window this way. But still, this is confusing - especially the fact that `destroy` signal isn't always emitted.

I propose the following solution:
- Unlink the file in `.save_session()`
- Add `.save_session()` call before `.close()`

samuelba: changed according to your proposal, at least I tried, not sure if my approach with the windows close signal is correct

ddterm/app/application.js
samuelba: removed this

ddterm/app/application.js
samuelba: removed this

ddterm/app/appwindow.js
```diff
@@ -333,8 +336,10 @@ class DDTermAppWindow extends Gtk.ApplicationWindow {
         this.update_show_shortcuts();
 
         this.connect('notify::is-empty', () => {
-            if (this.is_empty)
+            if (this.is_empty) {
+                this.emit('close');
```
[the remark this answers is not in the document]
samuelba: make sense, wasn't aware of this, changed it to call save_session() directly

samuelba:
> So now you don't `unlink()` the session file on any errors at all.
> 
> It may be fine, since it should be overwritten on the next correct shutdown - but it also would be nice to remove the file if it can't be parsed as `Variant` or when further restoring of the session (tabs) throws an error. But this is optional.

Ok, now I'm checking the data/variant and delete the session file if data is empty (= session file is empty) or variant is not normal.

ddterm/app/application.js
```diff
@@ -659,14 +659,26 @@ class Application extends Gtk.Application {
...
             );
 
+            if (!data_variant.is_normal_form()) {
+                logError(new Error('Invalid session file data variant, delete session file.'));
+                GLib.unlink(this.session_file_path);
+                return;
+            }
+
             this.ensure_window().deserialize_state(data_variant);
+        } else {
```
amezin, line 669: This `else` gets executed when the file is empty - not on error. And what would happen if something in `deserialize_state()` (or in functions/methods it calls) throws?

ddterm/app/application.js
```diff
@@ -659,14 +659,26 @@ class Application extends Gtk.Application {
                 false
             );
 
+            if (!data_variant.is_normal_form()) {
+                logError(new Error('Invalid session file data variant, delete session file.'));
+                GLib.unlink(this.session_file_path);
```
amezin, line 664: If you handle errors using `catch {}` on the top level, you could just `throw` an error here instead

samuelba: changed it to throw an error and try/catch the errors in `restore_session()`, I moved try/catch to  `restore_session()`, in case it is called from somewhere else in the future

ddterm/app/application.js
samuelba: removed the else

samuelba:
looks like your changes https://github.com/ddterm/gnome-shell-extension-ddterm/commit/dd6bfa26f059cb10ba85af1f9d08e7bc13d06f0a, broke the PR actions

image

PR #1209, opened 2025-02-25
[WIP] Gtk4 port

amezin:
https://github.com/ddterm/gnome-shell-extension-ddterm/issues/652

- Memory leaks
- 50% of WM tests fail
- Closing a tab/page doesn't kill the child process
- No transparent background

`WeakRef`s work correctly only starting from GJS 1.80/GNOME 46. Fix was backported to GNOME 45 branch, but most distros didn't update to GJS 1.78.5 patch. `GObject.connect_object` also available starting from GJS 1.80/GNOME 46 only. So I'm trying to break reference cycles in widget's `unrealize`.

I hate GJS.

### DIFF
--- a/ddterm/app/accellabel.js
+++ b/ddterm/app/accellabel.js
@@ -29,7 +29,7 @@ class DDTermAccelLabel extends Gtk.Label {
 
     #realize() {
         this.#hierarchy_handler =
-            this.connect('hierarchy-changed', this.#update_hierarchy.bind(this));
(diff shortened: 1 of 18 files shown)


amezin:
Maybe it's time to drop support for GNOME 45 and earlier.

amezin:
https://gitlab.gnome.org/GNOME/gtk/-/issues/5930

PR #1278, opened 2025-05-03
Don't restart terminated processes during session restore

amezin:
Don't restart processes that terminated before the session was saved.

Fixes https://github.com/ddterm/gnome-shell-extension-ddterm/issues/1277


amezin:
TODO: immediately after creating a new page, the child process isn't yet running, but session is saved at this moment.

Need to either:
- re-save when `child-pid`/running state changes
- treat the process as running

Or maybe even both

PR #1441, opened 2025-09-04
tests: fixes for GNOME 49

amezin:


andypost:
confirm the issue opening on the current 25.10

PR #1618, opened 2025-12-05
app: expect `GApplicationCommandLine.done()` to be available

amezin:
https://gitlab.gnome.org/GNOME/glib/-/issues/596


amezin:
Mutter v47 requires GLib 2.80.

But on GNOME v46 nothing really forces GLib to be new enough. But to get older GLib with GNOME 46 you'll have to build everything from source, no known distro provides that combination out of box, and even Gentoo doesn't have older GLib versions in the repository.

PR #1733, opened 2026-02-06
Korean translation of README.md

seuimi:
For the convenience of our Korean users, a Korean translation has been provided. Please refer to it for a smoother experience on GitHub.


seuimi:
@amezin 
On second thought, I realized that the app itself is already so well-translated that a full translation of the README.md might be unnecessary. The in-app localization is more than enough for Korean users to get started. Thanks again for your amazing support on the translation!

Issue #64, opened 2021-07-12
Copy on select feature


abdulrahman19:
First of all, I want to thank everyone who works on this awesome extension, but I think this extension missing a critical feature which is `copy on select` to copy any text to the clipboard when select them, it saves a lot of time, please add it!

amezin:
As far as I can tell, the feature is small and simple; somebody with basic Gtk and gjs knowledge could implement it.

amezin:
While working on a search bar, I've just discovered that Vte automatically copies selected text to the "primary" selection and pastes that text on a middle click.

So this is already implemented in Vte, and I see no reason to alter its behavior.

alejoar:
Doesn't work with tmux

PR #1778, opened 2026-02-23
Implement copy on selection

mikelei8291:
This feature was requested in #64, but was closed due to the fact that Vte automatically copies the selected text to the "primary" selection as you said in the [comment](https://github.com/ddterm/gnome-shell-extension-ddterm/issues/64#issuecomment-922246812).

However, the "primary" selection is only usable with the middle click paste action, not the usual <kbd>Ctrl</kbd>+<kbd>V</kbd> keyboard shortcut or any context menu entry that pastes content from the "clipboard" selection.

This PR adds an option to the settings of the extension under the "Text" section. When enabled, the selected text will also be copied to the "clipboard" selection, so it can be easily pasted using the keyboard shortcut or the context menu entry.

### DIFF
--- a/ddterm/app/terminal.js
+++ b/ddterm/app/terminal.js
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2023 Aleksandr Mezin <mezin.alexander@gmail.com>
 // SPDX-FileContributor: Jing Yen Loh
+// SPDX-FileContributor: Mike Lei
 //
 // SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -328,6 +329,13 @@ const TerminalBase = GObject.registerClass({
             GObject.ParamFlags.READABLE,
             true
         ),
+        'copy-on-selection': GObject.ParamSpec.boolean(
+            'copy-on-selection',
+            null,
+            null,
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.EXPLICIT_NOTIFY,
+            false
+        ),
     },
 }, class DDTermTerminal extends Vte.Terminal {
     _init(params) {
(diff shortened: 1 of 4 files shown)


ddterm/app/terminal.js
```diff
@@ -328,6 +329,13 @@ const TerminalBase = GObject.registerClass({
             GObject.ParamFlags.READABLE,
             true
         ),
+        'copy-on-selection': GObject.ParamSpec.boolean(
+            'copy-on-selection',
+            null,
+            null,
+            GObject.ParamFlags.READABLE,
```
amezin: No matter where the property is declared, it must be `READWRITE | EXPLICIT_NOTIFY` (as it needs to be writable, and GJS generates getters/setters that emit notifications)

ddterm/app/terminalpage.js
```diff
@@ -348,6 +349,9 @@ export const TerminalPage = GObject.registerClass({
         terminal_actions.add_action(copy_html_action);
 
         this.terminal.connect('selection-changed', () => {
+            if (this.terminal.get_has_selection() && this.terminal_settings.copy_on_selection) {
```
amezin: This can be implemented in the terminal widget entirely. Please move this to terminal.js.

ddterm/app/terminal.js
```diff
@@ -328,6 +329,13 @@ const TerminalBase = GObject.registerClass({
...
+            false
+        ),
```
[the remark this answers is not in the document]
amezin: No, `copy-on-selection` doesn't need to be removed. Instead, it should be writable, and the implementation should be moved into `TerminalBase` too.

amezin (changes requested):

amezin (approved):

PR #1833, opened 2026-03-21
Added feature to resize horizontally & apply XY position offset 

meepak:
image

### DIFF
--- a/ddterm/pref/positionsize.js
+++ b/ddterm/pref/positionsize.js
@@ -7,7 +7,7 @@ import GObject from 'gi://GObject';
 import Gio from 'gi://Gio';
 import Gtk from 'gi://Gtk';
 
-import { PreferencesGroup, ComboRow, ScaleRow } from './util.js';
+import { PreferencesGroup, ComboRow } from './util.js';
 import { Monitor } from '../util/displayconfig.js';
 
 class SpecialMonitor extends Monitor {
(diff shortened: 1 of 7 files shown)


schemas/com.github.amezin.ddterm.gschema.xml
```diff
@@ -144,6 +148,14 @@ SPDX-License-Identifier: GPL-3.0-or-later
     <key name="window-position" enum="com.github.amezin.ddterm.WindowPosition">
       <default>'top'</default>
     </key>
+    <key name="pos-x-offset" type="d">
+      <default>0.0</default>
+      <range min="-10000.0" max="10000.0"/>
```
amezin: To simplify things, let's start with centered (along the "secondary" axis) window

schemas/com.github.amezin.ddterm.gschema.xml
```diff
@@ -134,7 +134,11 @@ SPDX-License-Identifier: GPL-3.0-or-later
   </enum>
 
   <schema path="/com/github/amezin/ddterm/" id="com.github.amezin.ddterm">
-    <key name="window-size" type="d">
+    <key name="window-size-x" type="d">
```
amezin: Let's keep `window-size` setting, and add `window-size-secondary`. When position is top/bottom, `window-size` will control height, `window-size-secondary` will control width; and when position is left/right, `window-size` should control width, and `window-size-secondary` should control height. I.e. just keep the existing behavior for `window-size`, so, for example, keybinding actions will keep working.

amezin (changes requested):
Let's try to:
1) Reduce the amount of changes in this PR (keep `window-size` behavior, don't settings for offsets for now)
2) This change needs tests - and getting them to pass will, most likely, be the hardest part

schemas/com.github.amezin.ddterm.gschema.xml
meepak: Implemented as you suggested. I removed position offset from this branch. 
Some of the tests are failing, for me those are failing in master branch too. Maybe you can have a look and suggest something. Thanks.

ddterm/pref/positionsize.js
```diff
@@ -280,38 +280,24 @@ export class PositionSizeGroup extends PreferencesGroup {
...
-        this.add(window_size_row);
+        this.add_scale_row({
+            key: 'window-size-secondary',
+            adjustment:  new Gtk.Adjustment({
+                upper: 1,
+                step_increment: 0.01,
+                page_increment: 0.10,
+            }),
+            title: this.gettext('Window Size Secondary'),
+        });
```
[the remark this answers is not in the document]
amezin: Doesn't matter, this isn't even the final name for the option

schemas/com.github.amezin.ddterm.gschema.xml
amezin: > for me those are failing in master branch too

Could you tell me which tests exactly?

All tests pass on `master`, both in CI environment(s), and on my laptop

amezin:
[Alternative approach](https://github.com/ddterm/gnome-shell-extension-ddterm/issues/129#issuecomment-2870264295) is implemented in [v63.2.0](https://github.com/ddterm/gnome-shell-extension-ddterm/releases/tag/v63.2.0).

PR #1865, opened 2026-04-18
Add GNOME Shell 50 support

mauztor:
This PR adds support for GNOME Shell 50.

The extension previously used Meta.is_wayland_compositor() which was removed in GNOME Shell 50. This fix uses multiple fallbacks to detect Wayland session:

1. XDG_SESSION_TYPE environment variable
2. global.display.get_protocols() (if available)
3. Meta.get_backend().get_is_wayland() (if available)

Also adds Shell 50 to supported versions in metadata.json.


amezin:
Why do you send a PR that's obviously not correct, even the .json file isn't valid json

PR #1974, opened 2026-06-07
extension/dbusapi: use async i/o

amezin:
https://github.com/ddterm/gnome-shell-extension-ddterm/issues/1954

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,11 @@ The format is based on [Keep a Changelog], and this project adheres to
 ### Fixed
 
 - Replaced blocking i/o with async i/o in shell extension code:
-[#1954], [#1973].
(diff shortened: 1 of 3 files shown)


amezin:
@coderabbitai Is it possible that GNOME Shell will call `.disable()` while async `.enable()` is in progress?

amezin:
@coderabbitai

> 1. GNOME Shell calls DDTermExtension.enable(). Since it's async, it returns a Promise immediately and GNOME Shell moves on.
> GNOME Shell doesn't await the returned Promise, so this is likely silently swallowed.

Or does it?

https://github.com/GNOME/gnome-shell/blob/0cd0d774d91f9c2c0c129425f6c45fa228c04225/js/ui/extensionSystem.js#L265

PR #2023, opened 2026-06-23
Separate spacer background opacity slider.

raleik-pl:
This adds a separate slider in the preferences, allowing to set background opacity individually for the "work area" and for the "spacer".
image
image


amezin:
1. This PR shouldn't add new files to the repository
2. There should be a switch - whether spacer opacity should follow terminal opacity or not
3. Spacer opacity settings should be on "Window" page

raleik-pl:
Redundant file removed. 
Since "Transparent background" toggle and the "background opacity" slider are already in the "Terminal" tab (not in "Window") I figured this will be the best place possible to put just another slider next to the one existing. I think this makes it consistent and controllable with just two clicks to make the opacity match or not. 
Regards

amezin:
> Since "Transparent background" toggle and the "background opacity" slider are already in the "Terminal" tab (not in "Window") I figured this will be the best place possible to put just another slider next to the one existing. I think this makes it consistent and controllable with just two clicks to make the opacity match or not.

Spacers aren't part of the terminal, so the settings for them should be on "Window" page. This will become more important when profiles https://github.com/ddterm/gnome-shell-extension-ddterm/issues/638 will be implemented.

amezin:
I've moved dark mode setting to "Window" page for the same reason: https://github.com/ddterm/gnome-shell-extension-ddterm/pull/2027

Please move spacer opacity settings to the same "Appearance" group

amezin:
After thinking a bit more, it would make more sense to have "primary" opacity controls on the Window page:

Window -> Appearance -> Transparent Background (expander with a switch) -> Background Opacity (scale row)

I.e. to move current opacity controls to Window -> Appearance.

And on the Terminal page, have an override option:

Terminal -> Colors -> Override Background Opacity (expander with a switch) -> Background Opacity (scale row)

Which should affect only the terminal background, not spacers (if the override switch is on).

This approach won't need any behavior changes during profiles implementation, and seems straightforward to implement.

PR #2024, opened 2026-06-23
app/window: fix lost kbd focus on split pane close

van-riper:
Closing one pane of a split window causes the whole window to lose keyboard focus, meaning you have to click back on the window to focus it again.  This is caused by `_notebook_notify_n_pages` calling `grab_focus()` while HdyTabView was still removing the page. GTK then cleared focus once removal finished, so the window becomes unfocused.

Steps to reproduce:

1. Open ddterm
2. Split the window into two panes
3. Run `exit` in one of the panes
4. The remaining pane is unfocused

This fix defers the grab to a `GLib` idle callback, so it runs after the removal settles. A guard keeps it from being scheduled more than once, and a `destroy` handler removes any pending source.

Now, when running `exit` a split pane, focus is restored back to the remaining pane without needing to click on the window. Tested by hand and with `test_app.py` returned all green.

### DIFF
--- a/ddterm/app/appwindow.js
+++ b/ddterm/app/appwindow.js
@@ -256,7 +256,6 @@ export class AppWindow extends Gtk.ApplicationWindow {
     }
 
     #active_notebook = null;
-    #refocus_source = null;
 
     constructor(params) {
         super({
@@ -348,13 +347,6 @@ export class AppWindow extends Gtk.ApplicationWindow {
         });
 
         this.#setup_size_sync();
-
-        this.connect('destroy', () => {
-            if (this.#refocus_source) {
-                GLib.Source.remove(this.#refocus_source);
-                this.#refocus_source = null;
-            }
-        });
     }
 
     get hide_on_close() {
@@ -518,23 +510,10 @@ export class AppWindow extends Gtk.ApplicationWindow {
         this.notebook2.bind_settings(settings);
     }
 
-    _notebook_notify_n_pages(notebook) {
-        notebook.visible = notebook.n_pages > 0;
+    _notebook_notify_visible(notebook) {
+        if (!notebook.get_visible())
+            this.grab_focus();
 
-        if (notebook.get_visible() || this.#refocus_source)
-            return;
-
-        this.#refocus_source = GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
-            this.#refocus_source = null;
-
-            if (!this.in_destruction())
-                this.grab_focus();
-
-            return GLib.SOURCE_REMOVE;
-        });
-    }
-
-    _notebook_notify_visible() {
         this.freeze_notify();
 
         try {
--- a/ddterm/app/appwindow.js
+++ b/ddterm/app/appwindow.js
@@ -510,10 +510,14 @@ export class AppWindow extends Gtk.ApplicationWindow {
         this.notebook2.bind_settings(settings);
     }
 
-    _notebook_notify_visible(notebook) {
+    _notebook_notify_n_pages(notebook) {
+        notebook.visible = notebook.n_pages > 0;
+
         if (!notebook.get_visible())
             this.grab_focus();
+    }
 
+    _notebook_notify_visible() {
         this.freeze_notify();
 
         try {
--- a/ddterm/app/appwindow.js
+++ b/ddterm/app/appwindow.js
@@ -256,6 +256,7 @@ export class AppWindow extends Gtk.ApplicationWindow {
     }
 
     #active_notebook = null;
+    #refocus_source = null;
 
     constructor(params) {
         super({
@@ -347,6 +348,13 @@ export class AppWindow extends Gtk.ApplicationWindow {
         });
 
         this.#setup_size_sync();
+
+        this.connect('destroy', () => {
+            if (this.#refocus_source) {
+                GLib.Source.remove(this.#refocus_source);
+                this.#refocus_source = null;
+            }
+        });
     }
 
     get hide_on_close() {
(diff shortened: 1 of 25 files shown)


ddterm/app/appwindow.js
```diff
@@ -513,8 +521,17 @@ export class AppWindow extends Gtk.ApplicationWindow {
     _notebook_notify_n_pages(notebook) {
         notebook.visible = notebook.n_pages > 0;
 
-        if (!notebook.get_visible())
-            this.grab_focus();
+        if (notebook.get_visible() || this.#refocus_source)
+            return;
+
+        this.#refocus_source = GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
```
amezin: There should be a solution without idle_add.

Which widget currently receives focus (without these changes)?

van-riper: None to my knowledge since `window.get_focus()` returns null after the pane closes. The closing terminal keeps focus until it's destroyed, and GtkWindow resets focus to null right after the synchronous `grab_focus()` runs.

amezin: What if you call `window.grab_focus()` after a split pane/notebook got hidden (in `notify::visible`)?

amezin: Oh, wait, it's already done

van-riper: Would you still prefer a different approach to this?

amezin: Why doesn't [`grab_focus()` call here](https://github.com/ddterm/gnome-shell-extension-ddterm/blob/be55e22c5674b48715c0c62609ff22bdfc726c8b/ddterm/app/appwindow.js#L517) solve the issue?

amezin: Maybe connecting the signal handler with "after" flag could help?

van-riper: Yeah that makes sense. The grab runs before HdyTabView finishes removing the page, so it should get overwritten. I'll try the after flag and drop the idle_add.

amezin: TabView [emits `SIGNAL_PAGE_DETACHED`](https://github.com/GNOME/libhandy/blob/9e72e732861fb6f58c0f69f9a1b89ea83843a05f/src/hdy-tab-view.c#L764) after removing the page from the container. So maybe the Notebook class should rely on that signal instead of `notify::n-pages` (there are already handlers for attached/detached signals).

It doesn't look like `after` will help.

van-riper: Yeah it's not retaining focus with the after flag. I'll see if using the detached handler instead does any better

van-riper:
The AlmaLinux wayland tests are failing sequentially in CI, do you suspect why?

amezin:
> The AlmaLinux wayland tests are failing sequentially in CI, do you suspect why?

It looks like sometimes the app locks up, somewhere in Gio/GtkApplication. Never seen it happen during normal use, but in the tests it sometimes happens.

It's unrelated to this pull request, happens on master too

ddterm/app/notebook.js
```diff
@@ -433,6 +433,7 @@ export class Notebook extends Gtk.Box {
                 bindings.pop().unbind();
         });
 
+        this.visible = this.n_pages > 0;
```
amezin: Usually Gtk widgets don't change their own `visible` property, and I'd prefer to follow this rule.

Maybe instead change how/when `notify::n-pages` is emitted?

Both logically, and from looking at HdyTabView sources, it should [be emitted](https://github.com/GNOME/libhandy/blob/9e72e732861fb6f58c0f69f9a1b89ea83843a05f/src/hdy-tab-view.c#L625) only when a page is [attached](https://github.com/GNOME/libhandy/blob/9e72e732861fb6f58c0f69f9a1b89ea83843a05f/src/hdy-tab-view.c#L677) or [detached](https://github.com/GNOME/libhandy/blob/9e72e732861fb6f58c0f69f9a1b89ea83843a05f/src/hdy-tab-view.c#L752).

amezin (changes requested):

ddterm/app/notebook.js
van-riper: @amezin markdownlint failed in CI even though everything is clean, can you rerun the job?

ddterm/app/notebook.js
```diff
@@ -437,6 +437,7 @@ export class Notebook extends Gtk.Box {
         this.grab_focus();
         this.#update_tab_switch_accels();
         this.emit('session-update');
+        this.notify('n-pages');
```
amezin: It's likely better to emit/notify at the places where you updated `visible` previously. In this case - before `grab_focus()`, because visibility should be updated before the call

ddterm/app/notebook.js
```diff
@@ -450,6 +451,7 @@ export class Notebook extends Gtk.Box {
 
         this.#update_tab_switch_accels();
         this.emit('session-update');
+        this.notify('n-pages');
```
amezin: Same here - I'd put this line before `#update_tab_switch_accels()`, for consistency with `_page_attached`, even though it likely doesn't matter here

amezin (changes requested):

ddterm/app/notebook.js
amezin: It's not markdownlint, it's GitHub API, it decided to return a HTML page instead of a list of releases in JSON

ddterm/app/notebook.js
```diff
@@ -433,6 +433,7 @@ export class Notebook extends Gtk.Box {
                 bindings.pop().unbind();
         });
 
+        this.notify('n-pages');
```
amezin, line 440: One last thing: could you add a comment describing the issue?

I.e. when I look at this code without the context, my first question would be "why isn't it just something like `view.connect('notify::n-pages', () => notify('n-pages'))`.

amezin (changes requested):

amezin (approved):

PR #2069, opened 2026-07-19
app: check GJS version and warn

amezin:
https://github.com/ddterm/gnome-shell-extension-ddterm/issues/1624


amezin:
Need to detect this message in tests

PR #2092, opened 2026-07-31
npm: switch back to git install of `eslint-config-gnome`

amezin:
- npm v12 switches `allow-remote` config from `all` to `none` by default (this breaks install of `eslint-config-gnome`).

- On Ubuntu 26.04 Actions runner images, npm 11.16.0 is installed. Hopefully, it has all git dependency bugs fixed.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -6,5 +6,6 @@ git-tag-version = false
 preid = dev
 ignore-scripts = true
 allow-git = root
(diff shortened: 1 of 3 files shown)


amezin:
It seems that CodeRabbit still has issues with git deps, and npm still does some nonsense with "integrity"

PR #2109, opened 2026-08-16
Enable GNOME 51 support

amezin:


amezin:
https://gitlab.gnome.org/GNOME/mutter/-/work_items/5010

PR #2120, opened 2026-09-12
Fix typos

kianmeng:
Found via `codespell -S po,package-lock.json -L filetest`

### DIFF
--- a/tests/dbus-interfaces/org.gnome.Mutter.DisplayConfig.xml
+++ b/tests/dbus-interfaces/org.gnome.Mutter.DisplayConfig.xml
@@ -273,7 +273,7 @@
 	- -1: unknown (unsupported)
 
         A client should not attempt to change the powersave mode
-	from -1 (unknown) to any other value, and viceversa.
+	from -1 (unknown) to any other value, and vice-versa.
 	Note that the actual effects of the different values
 	depend on the hardware and the kernel driver in use, and
 	it's perfectly possible that all values different than on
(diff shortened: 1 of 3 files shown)


tests/dbus-interfaces/org.gnome.Mutter.DisplayConfig.xml
```diff
@@ -273,7 +273,7 @@
 	- -1: unknown (unsupported)
 
         A client should not attempt to change the powersave mode
-	from -1 (unknown) to any other value, and viceversa.
+	from -1 (unknown) to any other value, and vice-versa.
```
amezin: This file is copied from https://gitlab.gnome.org/GNOME/mutter/-/blob/main/data/dbus-interfaces/org.gnome.Mutter.DisplayConfig.xml?ref_type=heads

If you want to fix a typo in this file, you should fix it there

tests/dbus-interfaces/org.gnome.Mutter.DisplayConfig.xml
amezin, line 1: This file is copied as is from https://gitlab.gnome.org/GNOME/mutter/-/blob/main/data/dbus-interfaces/org.gnome.Mutter.DisplayConfig.xml

If you want to fix a typo, you should do it there

amezin (changes requested):
Please add codespell to CI, and exclude `test/dbus-interfaces/org.gnome.*`, `tools/heapgraph*`, `tools/heapdot*` from the check. These are files copied from other projects, and I'd prefer to keep them unmodified

PR #2129, opened 2026-09-17
fix: allow Shift+arrows as keyboard shortcut bindings

### LINKED ISSUE #2121: Shift+arrows not allowed as a keyboard shortcut binding
### GNOME Shell version

GNOME Shell 50.1

### Linux distro

Ubuntu 26.04 LTS

### ddterm version

ddterm 63.2.3 4f64fbe89 (72)
Extension 63.2.3 4f64fbe89 (72)

### Session type

Wayland (`echo $WAYLAND_DISPLAY` output is non-empty)

### All installed extensions

```Text

```

### ddterm settings dump

```ini

```

### Description

Hi,
I'd like to use Shift+Left & Shift+Right (i.e. arrow keys with just shift) to move between tabs. Unfortunately, I can't set it in preferences. I believe it's rejected by [this function](https://github.com/ddterm/gnome-shell-extension-ddterm/blob/master/ddterm/pref/shortcuts.js#L116). Is this intended? I think using these bindings doesn't break anything, and when I forced them via dconf-editor it works just fine. 

Maybe the logic should be `(FORBIDDEN_KEYVALS.includes(keyval) && !mask)` in the last line of is_valid_binding?

Thanks!

niukanen1:
Fixes #2121

The validation in `is_valid_binding` rejected every keyval from `FORBIDDEN_KEYVALS` regardless of modifiers, so Shift+Left/Up/Right/Down could not be bound in the preferences dialog, even though the bindings work fine when set via dconf

Arrow keys with just the Shift modifier are now allowed through a separate `ARROW_KEYVALS` check. Everything else keeps the existing behavior — Shift+Tab and other forbidden keyvals stay invalid, per your note in the issue

Syntax checked with `node --check`; no unit-test harness exists for `is_valid_binding` (the `ddterm/pref/test/` GJS harness needs a live GTK session)


amezin:
How does it work when a text entry widget is focused? For example, search bar (<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd>) - <kbd>Ctrl</kbd>+<kbd>Left</kbd> and <kbd>Ctrl</kbd>+<kbd>Right</kbd> should select (or deselect) entered characters

niukanen1:
Checked the propagation path. Application accelerators are handled by GtkWindow during the capture phase — before the key event reaches the focused widget. So yes: if a user assigns Shift+Left as a ddterm shortcut, that binding takes the key and the focused GtkEntry (search bar) never sees it — text selection with Shift+arrows stops working while that entry has focus.

That's the same trade-off any custom binding on a modifier combo already makes today (e.g. assigning Ctrl+L or Ctrl+Shift+F to a ddterm action would shadow the search toggle — the search bar itself only opens via Ctrl+Shift+F). The PR doesn't change this: Shift+arrows remain opt-in — no default binding uses them, so out of the box nothing is affected.

If shadowing the search entry is a concern, one containment option is to keep bare Shift+arrows valid only for the window-level actions (toggle/hide) and not for terminal-scoped ones — but that split doesn't exist in the current keybinding model, so I'd rather keep the validator simple unless you want that distinction.

amezin:
So this breaks normal text editing

Also I don't like talking to clankers disguised as humans

Closing

